PR #7, opened 2025-10-07
Create channels and types for them

### LINKED ISSUE #2: Setting up channnels for tokio task communication
Current version has a total of three tokio tasks per client (single tcp connection). Also, we decided also to spawn tasks for every file, we would process. So, we should set up an effective way of data sharing between them.

MikhailSvirukov:
- create special types to use with mpsc channels
- create wrappers for them
- change unbounded_channel with channel 

Resolves #2


src/message_types.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Response container with transaction ID and procedure result
+pub struct Reply {
+    pub xid: u32,
+    pub result: ProcResult,
+}
+
+/// Result of procedure execution - either success or error
+pub enum ProcResult {
```
IgorErin, line 26: Single file `proc.rs`. We should use rust namespaces (proc::Result) --- not C style names

src/stream_writer.rs
```diff
@@ -0,0 +1,54 @@
+#![allow(dead_code)]
+use tokio::net::tcp::OwnedWriteHalf;
+use tokio::sync::mpsc::Receiver;
+use tokio::task::JoinHandle;
+
+use crate::message_types::{EarlyReply, EarlyResult, ProcResult, Reply};
+
+/// Writes [`vfs_task::VfsTask`] responses to a network connection.
+pub struct StreamWriter {
```
IgorErin, line 9: Why stream writer? Why ReadTask still ReadTask, not StreamRead task? For what `Stream` stands for?

IgorErin (changes requested):
For what Stream?

src/message_types.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Represents a procedure call with transaction ID and command arguments
+pub struct Procedure {
+    xid: u32,
+    args: Command,
+}
+
+/// Enum of supported protocol commands
+pub enum Command {
+    NFSv3Command,
+    MountCommand,
```
IgorErin, line 16: What use case to extract mount from Nfs3?

src/message_types.rs
IgorErin, line 26: Use default Result for Result

src/message_types.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Success response types for different protocol operations
+pub enum ProcResOk {
+    NFSv3ResOk,
+    MountResok,
+}
+
+/// Early reply sent before full procedure completion
+pub struct EarlyReply {
```
IgorErin, line 44: is it rpc error or auth error?

src/lib.rs
```diff
@@ -1,16 +1,19 @@
...
 mod read_task;
+mod stream_writer;
 mod vfs_task;
-mod write_task;
 
 use tokio::net::TcpListener;
 use tokio::net::TcpStream;
-use tokio::sync::mpsc;
 
+use crate::message_types::{create_early_reply_channel, create_proc_channel, create_reply_channel};
```
IgorErin, line 11: is it formatted?

IgorErin:
Why you choose tokio channel?

IgorErin:
We don't have to public Task structure. Let's create one function

```
let vfs_task = vfs_task::spawn()
```

vfs_task is a view to VfsTask backed by channel

src/stream_writer.rs
MikhailSvirukov: Well, i couldn't think of anything better) I mean writing sequence of bytes (as stream) into socket, but i didn't want to use word "socket"

src/message_types.rs
MikhailSvirukov: It is unified type, that represents correct arguments (after parsing and auth), that can be processed by VfsTask

src/message_types.rs
MikhailSvirukov: Auth error is one of RPC errors.
But in general, it is a unified type for things, that we are going to process without using VfsTask

IgorErin:
FYI: We can't tests this without parser for commands

MikhailSvirukov:
> Why you choose tokio channel?

They are non-blocking, and we need to limit memory usage by channels (allocator would protect from too large requests and backpressure - from big amount of small ones)

src/message_types.rs
MikhailSvirukov: They would have types later. Probably, we should write all types and structures from rfc first)

IgorErin: What types? Is it RFC error or our own?

src/message_types.rs
IgorErin: I see, but why Mount is distinct? Let's write just empty `enum` for now

src/stream_writer.rs
```diff
@@ -0,0 +1,54 @@
...
+    writehalf: OwnedWriteHalf,
+    reply_recv: Receiver<Reply>,
+    early_recv: Receiver<EarlyReply>,
+}
+
+impl StreamWriter {
+    /// Creates new instance of [`StreamWriter`]
```
IgorErin, line 16:
```suggestion
    /// Creates new instance of [`StreamWriter`].
```

src/stream_writer.rs
IgorErin: StreamWriter sounds like structure that will write raw bytes, but this task writes structured NFS/RPC responses.  It's not a stream since we write responses without order and in delayed fashion (think about caching of NFS3 writes inside VFS pool) then ResponseWrite? But why not just WriteTask?

src/stream_writer.rs
```diff
@@ -0,0 +1,54 @@
...
+    ) -> JoinHandle<()> {
+        tokio::spawn(Self { writehalf, reply_recv, early_recv }.run())
+    }
+
+    async fn run(mut self) {
+        loop {
+            tokio::select! {
+                Some(early_reply) = self.early_recv.recv() => {
+                    match early_reply.result {
+                        EarlyResult::RPCError => {
```
IgorErin, line 30: Why distinct types? Why not std::Result?

src/message_types.rs
IgorErin: From spec:

```
It is important that this procedure do no work at all so
      that it can be used to measure the overhead of processing
      a service request.
```

But if we use separate channel we would measure only early responses not Vfs buffer delay

MikhailSvirukov: Okay, that is good point

VadimYarovoy:
@MikhailSvirukov update on main pls, and what is status for now?

PR #8, opened 2025-10-08
Add Vfs trait for NFS server

### LINKED ISSUE #4: Vfs trait
We should write Vfs trait for nfs3 file operations. See RFC 1813.

VadimYarovoy:
Resolves #4

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,29 +1,35 @@
 //! NFS Mamont - A Network File System (NFS) server implementation in Rust.
 
 mod read_task;
+pub mod vfs;
 mod vfs_task;
 mod write_task;
 
+use std::sync::Arc;
 use tokio::net::TcpListener;
 use tokio::net::TcpStream;
 use tokio::sync::mpsc;
 
 use crate::read_task::ReadTask;
+use crate::vfs::Vfs;
 use crate::vfs_task::VfsTask;
 use crate::write_task::WriteTask;
 
 /// Starts the NFS server and processes client connections.
-pub async fn handle_forever(listener: TcpListener) -> std::io::Result<()> {
+pub async fn handle_forever<V: Vfs + 'static>(
+    listener: TcpListener,
+    vfs: Arc<V>,
+) -> std::io::Result<()> {
     loop {
         let (socket, _) = listener.accept().await?;
 
         socket.set_nodelay(true)?;
 
-        process_socket(socket).await;
+        process_socket(socket, Arc::clone(&vfs)).await;
     }
 }
 
-async fn process_socket(socket: TcpStream) {
+async fn process_socket<V: Vfs + 'static>(socket: TcpStream, vfs: Arc<V>) {
     let (readhalf, writehalf) = socket.into_split();
     // channel for result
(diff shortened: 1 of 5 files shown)


src/vfs_task.rs
```diff
@@ -1,18 +1,27 @@
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+use std::sync::Arc;
```
IgorErin: import order

src/vfs_task.rs
```diff
@@ -1,18 +1,27 @@
...
 
-/// Process RPC commands,sends operation results to [`crate::write_task::WriteTask`].
-pub struct VfsTask {
+use crate::vfs::Vfs;
+
+/// Process RPC commands, sends operation results to [`crate::write_task::WriteTask`].
+pub struct VfsTask<V: Vfs + 'static> {
     _command_receiver: UnboundedReceiver<()>,
     _result_sender: UnboundedSender<()>,
+    _vfs: Arc<V>,
```
IgorErin: why Arc?

src/lib.rs
```diff
@@ -1,29 +1,35 @@
...
 use crate::read_task::ReadTask;
+use crate::vfs::Vfs;
 use crate::vfs_task::VfsTask;
 use crate::write_task::WriteTask;
 
 /// Starts the NFS server and processes client connections.
-pub async fn handle_forever(listener: TcpListener) -> std::io::Result<()> {
+pub async fn handle_forever<V: Vfs + 'static>(
+    listener: TcpListener,
+    vfs: Arc<V>,
```
IgorErin: Why arc?

IgorErin (changes requested):
Why Arc? Not all users want to use Arc. 

Mb add methods of Vfs trait rather then try to add it to VfsTask?

VadimYarovoy:
> Why Arc? Not all users want to use Arc.
> 
> Mb add methods of Vfs trait rather then try to add it to VfsTask?

sure, will remove it from Task and will add methods firstly

VadimYarovoy:
@MikhailSvirukov @IgorErin review as you will have time for it

src/vfs.rs
```diff
@@ -0,0 +1,499 @@
...
+    pub attr: Option<FileAttr>,
+}
+
+/// Result of READDIR (RFC 1813 3.3.16).
+#[derive(Debug, Clone, PartialEq)]
+pub struct ReadDirResult {
+    pub directory_attr: Option<FileAttr>,
+    pub cookie_verifier: CookieVerifier,
+    pub entries: Vec<DirectoryEntry>,
+    pub eof: bool,
```
MikhailSvirukov: Why do we need `eof` field?

VadimYarovoy: i had idea that there will be several read replies not to have one `ReadResult` on 1 TB file in Vec<u8>, probably need to fix interafece in future, for now it must be as simple as possible, can remove it or jsut left them unused

VadimYarovoy: removed - for now our goal to be more simple

MikhailSvirukov (approved):

IgorErin (approved):

PR #10, opened 2025-10-10
RFC types

MikhailSvirukov:
add types, defined by RPC message protocol and by rfc 1813 for NFSv3 and Mount protocols

### DIFF
--- a/src/mount.rs
+++ b/src/mount.rs
@@ -38,6 +38,7 @@ enum MountProgram {
     MOUNTPROC3_EXPORT = 5,
 }
 
+#[derive(Debug)]
 struct mountres3_ok {
     fhandle: fhandle3,
     auth_flavors: Vec<i32>,
@@ -45,6 +46,7 @@ struct mountres3_ok {
 
 type mountlist = Option<Box<mountbody>>;
 
+#[derive(Debug)]
 struct mountbody {
     ml_hostname: name,
     ml_directory: dirpath,
@@ -53,6 +55,7 @@ struct mountbody {
 
 type groups = Option<Box<groupnode>>;
 
+#[derive(Debug)]
 struct groupnode {
     gr_name: name,
     gr_next: groups,
@@ -60,6 +63,7 @@ struct groupnode {
 
 type exports = Option<Box<exportnode>>;
 
+#[derive(Debug)]
 struct exportnode {
     ex_dir: dirpath,
     ex_groups: groups,
--- a/src/mount.rs
+++ b/src/mount.rs
@@ -1,0 +1,67 @@
+#![allow(dead_code)]
+#![allow(non_camel_case_types, clippy::upper_case_acronyms)]
+
+const MOUNT_PROGRAM: u32 = 100005;
+const MOUNT_VERSION: u32 = 3;
+
+const MNTPATHLEN: u32 = 1024;
+const MNTNAMLEN: u32 = 255;
+const FHSIZE3: u32 = 64;
+
+type fhandle3 = Vec<u8>;
+type dirpath = Vec<u8>;
+type name = Vec<u8>;
+
+#[derive(Debug)]
+#[repr(u32)]
+enum mountstat3 {
+    MNT3_OK = 0,
+    MNT3ERR_PERM = 1,
+    MNT3ERR_NOENT = 2,
+    MNT3ERR_IO = 5,
+    MNT3ERR_ACCES = 13,
+    MNT3ERR_NOTDIR = 20,
+    MNT3ERR_INVAL = 22,
+    MNT3ERR_NAMETOOLONG = 63,
+    MNT3ERR_NOTSUPP = 10004,
+    MNT3ERR_SERVERFAULT = 10006,
+}
+
+#[derive(Debug)]
+#[repr(u32)]
+enum MountProgram {
+    MOUNTPROC3_NULL = 0,
+    MOUNTPROC3_MNT = 1,
+    MOUNTPROC3_DUMP = 2,
+    MOUNTPROC3_UMNT = 3,
+    MOUNTPROC3_UMNTALL = 4,
+    MOUNTPROC3_EXPORT = 5,
+}
+
+struct mountres3_ok {
+    fhandle: fhandle3,
+    auth_flavors: Vec<i32>,
+}
+
+type mountlist = Option<Box<mountbody>>;
+
+struct mountbody {
+    ml_hostname: name,
+    ml_directory: dirpath,
+    ml_next: mountlist,
+}
+
+type groups = Option<Box<groupnode>>;
+
+struct groupnode {
+    gr_name: name,
+    gr_next: groups,
+}
+
+type exports = Option<Box<exportnode>>;
+
+struct exportnode {
+    ex_dir: dirpath,
+    ex_groups: groups,
+    ex_next: exports,
+}
--- a/src/mount.rs
+++ b/src/mount.rs
@@ -1,19 +1,21 @@
-#![allow(dead_code)]
 #![allow(non_camel_case_types, clippy::upper_case_acronyms)]
 
+#[allow(dead_code)]
 const MOUNT_PROGRAM: u32 = 100005;
+#[allow(dead_code)]
 const MOUNT_VERSION: u32 = 3;
-
+#[allow(dead_code)]
 const MNTPATHLEN: u32 = 1024;
+#[allow(dead_code)]
 const MNTNAMLEN: u32 = 255;
+#[allow(dead_code)]
 const FHSIZE3: u32 = 64;
 
 type fhandle3 = Vec<u8>;
 type dirpath = Vec<u8>;
 type name = Vec<u8>;
 
-#[derive(Debug)]
-#[repr(u32)]
+#[allow(dead_code)]
 enum mountstat3 {
     MNT3_OK = 0,
     MNT3ERR_PERM = 1,
@@ -27,18 +29,18 @@ enum mountstat3 {
     MNT3ERR_SERVERFAULT = 10006,
 }
 
-#[derive(Debug)]
+#[allow(dead_code)]
 #[repr(u32)]
 enum MountProgram {
     MOUNTPROC3_NULL = 0,
-    MOUNTPROC3_MNT = 1,
+    MOUNTPROC3_MNT(dirpath) = 1,
     MOUNTPROC3_DUMP = 2,
-    MOUNTPROC3_UMNT = 3,
+    MOUNTPROC3_UMNT(dirpath) = 3,
     MOUNTPROC3_UMNTALL = 4,
     MOUNTPROC3_EXPORT = 5,
 }
 
-#[derive(Debug)]
+#[allow(dead_code)]
 struct mountres3_ok {
     fhandle: fhandle3,
     auth_flavors: Vec<i32>,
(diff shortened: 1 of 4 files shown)


MikhailSvirukov:
@IgorErin what do you think about names? Do we leave them in C-style or not?

MikhailSvirukov:
Also, are we planning to make them private (do we need any docs - for now i just copy discribtion of errors from rfc) and create some wrapper types?

src/nfsv3.rs
```diff
@@ -0,0 +1,711 @@
...
+const NFS_PROGRAM: u32 = 100003;
+const NFS_VERSION: u32 = 3;
+const NFS3_FHSIZE: u32 = 64;
+const NFS3_COOKIEVERFSIZE: u32 = 8;
+const NFS3_CREATEVERFSIZE: u32 = 8;
+const NFS3_WRITEVERFSIZE: u32 = 8;
+
+enum NFSProgram {
+    NFSPROC3_NULL = 0,
+    NFSPROC3_GETATTR = 1,
```
VadimYarovoy: why here not something like that?
```
 NFSPROC3_GETATTR(GETATTR3args) = 1,
```

cc @IgorErin

VadimYarovoy: i thought this was general idea of all these enums

MikhailSvirukov: Good question. My main concern was to write as close to specification as possible. I have united some C-enum with C-union, if they are used only together. Probably we are going to use enum separately. I will check again

src/nfsv3.rs
```diff
@@ -0,0 +1,711 @@
...
+#![allow(non_camel_case_types, clippy::upper_case_acronyms)]
+
+const NFS_PROGRAM: u32 = 100003;
+const NFS_VERSION: u32 = 3;
+const NFS3_FHSIZE: u32 = 64;
+const NFS3_COOKIEVERFSIZE: u32 = 8;
+const NFS3_CREATEVERFSIZE: u32 = 8;
+const NFS3_WRITEVERFSIZE: u32 = 8;
+
+enum NFSProgram {
```
VadimYarovoy: Maybe style from RFC, because enum name is in rust style and its fields are in RFC style

src/mount.rs
```diff
@@ -0,0 +1,71 @@
+#![allow(dead_code)]
+#![allow(non_camel_case_types, clippy::upper_case_acronyms)]
```
VadimYarovoy, line 1: better not use #![allow(...)] on whole file, use #[allow(...)] on each unused struct - it will be more simple to us in further development

VadimYarovoy:
@MikhailSvirukov @dsemak @IgorErin sounds like it is better to create Encoder/DEcoder firstly and then add structs (consts are ok in this PR) one by one with tests. Opinions?

MikhailSvirukov:
> @MikhailSvirukov @dsemak @IgorErin sounds like it is better to create Encoder/DEcoder firstly and then add structs (consts are ok in this PR) one by one with tests. Opinions?

Since we brought encoding/decoding up, i'd like to hear your opinion about libraries, i found: [nom](https://github.com/rust-bakery/nom) and it's serialization analogue - [cookie-factory](analogue). 
As for the first one - it is  actually parser-combinators lib, but apparently also used for net protocols parsing. 
They both has parsers/serializators for some basic types of XDR. 

As for the order - i don't see problem having them before implementing Decoder-parser. Though it might be better to them separately if we would go with two custom traits. But with parser it can be tricky to split parser in several parts.

So, opinions? @VadimYarovoy @IgorErin @dsemak

MikhailSvirukov:
They both go with MIT license, so i think it is okay for us

src/rpc.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+#[derive(Debug)]
+struct opaque_auth {
+    flavor: auth_flavor,
+    body: Vec<u8>,
+}
+
+#[derive(Debug)]
+#[repr(u32)]
```
IgorErin, line 95: no repr

src/mount.rs
```diff
@@ -0,0 +1,71 @@
...
+const MNTPATHLEN: u32 = 1024;
+const MNTNAMLEN: u32 = 255;
+const FHSIZE3: u32 = 64;
+
+type fhandle3 = Vec<u8>;
+type dirpath = Vec<u8>;
+type name = Vec<u8>;
+
+#[derive(Debug)]
+#[repr(u32)]
```
IgorErin: no repr

src/nfsv3.rs
IgorErin: A agree with you

src/rpc.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[allow(dead_code)]
+const RPC_VERSION: u32 = 2;
+
+#[allow(dead_code)]
+#[repr(u32)]
+enum accept_stat {
+    SUCCESS = 0,
+    PROG_UNAVAIL = 1,
+    PROG_MISMATTCH(mismatch_info) = 2,
```
VadimYarovoy:
```suggestion
    PROG_MISMATCH(mismatch_info) = 2,
```

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+#[allow(dead_code)]
+const NFS3_COOKIEVERFSIZE: u32 = 8;
+#[allow(dead_code)]
+const NFS3_CREATEVERFSIZE: u32 = 8;
+#[allow(dead_code)]
+const NFS3_WRITEVERFSIZE: u32 = 8;
+
+#[allow(dead_code)]
+#[repr(u32)]
+enum NFS_V3_PROGRAM {
```
VadimYarovoy: from RFC

```
program NFS_PROGRAM {
         version NFS_V3 {

            void
             NFSPROC3_NULL(void)                    = 0;

            GETATTR3res
             NFSPROC3_GETATTR(GETATTR3args)         = 1;
```

```suggestion
enum NFS_V3 {
```

VadimYarovoy (commented):
Naming issues

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+}
+
+#[allow(dead_code)]
+type sattrguard3 = Option<nfstime3>;
+
+#[allow(dead_code)]
+struct SETATTR3args {
+    object: nfs_fh3,
+    new_attribute: sattr3,
+    guard: Option<nfstime3>,
```
VadimYarovoy:
```suggestion
    guard: sattrguard3,
```

VadimYarovoy (dismissed):

VadimYarovoy:
@IgorErin @dsemak review as fast as possible

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+const NFS3_FHSIZE: u32 = 64;
+#[allow(dead_code)]
+const NFS3_COOKIEVERFSIZE: u32 = 8;
+#[allow(dead_code)]
+const NFS3_CREATEVERFSIZE: u32 = 8;
+#[allow(dead_code)]
+const NFS3_WRITEVERFSIZE: u32 = 8;
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 17:
```suggestion
```

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+}
+
+type post_op_fh3 = Option<nfs_fh3>;
+type set_mode3 = Option<mode3>;
+type set_uid3 = Option<uid3>;
+type set_gid3 = Option<gid3>;
+type set_size3 = Option<size3>;
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 223:
```suggestion
```

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+enum set_atime {
+    DONT_CHANGE = 0,
+    SET_TO_SERVER_TIME = 1,
+    SET_TO_CLIENT_TIME(nfstime3) = 2,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 231:
```suggestion
```

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+    verf: writeverf3,
+}
+
+#[allow(dead_code)]
+struct WRITE3resfail {
+    file_wcc: wcc_data,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 397:
```suggestion
```

src/nfsv3.rs
```diff
@@ -0,0 +1,730 @@
...
+}
+
+#[allow(dead_code)]
+struct devicedata3 {
+    dev_attributes: sattr3,
+    spec: specdata3,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 471:
```suggestion
```

src/rpc.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[allow(dead_code)]
+struct rpc_msg {
+    xid: u32,
+    body: rpc_body,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 43:
```suggestion
```

src/rpc.rs
```diff
@@ -0,0 +1,99 @@
...
+    rpcvers: u32,
+    prog: u32,
+    vers: u32,
+    proc: u32,
+    cred: opaque_auth,
+    verf: opaque_auth,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 60:
```suggestion
```

src/rpc.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[allow(dead_code)]
+struct opaque_auth {
+    flavor: auth_flavor,
+    body: Vec<u8>,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 95:
```suggestion
```

src/rpc.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[allow(dead_code)]
+struct accepted_reply {
+    verf: opaque_auth,
+    reply_data: accept_stat,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 79:
```suggestion
```

src/rpc.rs
```diff
@@ -0,0 +1,99 @@
+#![allow(non_camel_case_types, clippy::upper_case_acronyms)]
+
+#[allow(dead_code)]
+const RPC_VERSION: u32 = 2;
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 7:
```suggestion
```

src/mount.rs
```diff
@@ -0,0 +1,73 @@
...
+    MNT3ERR_ACCES = 13,
+    MNT3ERR_NOTDIR = 20,
+    MNT3ERR_INVAL = 22,
+    MNT3ERR_NAMETOOLONG = 63,
+    MNT3ERR_NOTSUPP = 10004,
+    MNT3ERR_SERVERFAULT = 10006,
+}
+
+#[allow(dead_code)]
+#[repr(u32)]
```
IgorErin, line 33:
```suggestion
```

IgorErin (changes requested):

src/rpc.rs
MikhailSvirukov: I receive [this](https://doc.rust-lang.org/error_codes/E0732.html) error

MikhailSvirukov: @IgorErin, or this one shouldn't normally happen?

VadimYarovoy (approved):

IgorErin (approved):

PR #11, opened 2025-10-14
feat: add async allocator

### LINKED ISSUE #5: Implement async allocator

IgorErin:
Resolves: #5


src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+        (as_mut, next)
+    }
+
+    fn repr_c_layout(fields: &[Layout]) -> Result<Layout, LayoutError> {
+        let mut layout = Layout::from_size_align(0, 1)?;
+        for &field in fields {
+            let (new_layout, _) = layout.extend(field)?;
+            layout = new_layout;
+        }
+        Ok(layout.pad_to_align())
```
IgorErin: No pad_to_align here

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    fn as_ptr(&self) -> *const BufferLayout {
+        self.0.as_ptr()
+    }
+}
+
+type Sender = mpsc::Sender<Buffer>;
+type Reciever = mpsc::Receiver<Buffer>;
+
+struct BufferChain {
```
IgorErin: Implement Cursor

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+        Ok(layout.pad_to_align())
+    }
+
+    fn compute_layout(layout_size: usize) -> Result<Layout, LayoutError> {
+        let header_layout = Self::repr_c_layout(Self::HEADER_BUFFER_FIELDS)?;
+        let (layout, _) =
+            header_layout.extend(Layout::for_value(vec![0u8; layout_size].as_slice()))?;
+
+        Ok(layout.pad_to_align())
+    }
```
VadimYarovoy: use array instead of vector

```suggestion
    fn compute_layout(payload_size: usize) -> Result<Layout, LayoutError> {
        let header_layout = Self::repr_c_layout(Self::HEADER_BUFFER_FIELDS)?;

        let payload_layout = Layout::array::<u8>(payload_size)?;
        let (full_layout, _) = header_layout.extend(payload_layout)?;

        Ok(full_layout.pad_to_align())
    }
```

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+        unsafe { (*self.0.as_ptr()).layout }
+    }
+
+    fn as_ptr(&self) -> *const BufferLayout {
+        self.0.as_ptr()
+    }
+}
+
+type Sender = mpsc::Sender<Buffer>;
+type Reciever = mpsc::Receiver<Buffer>;
```
VadimYarovoy:
```suggestion
type Receiver = mpsc::Receiver<Buffer>;
```

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    pub fn blocking_dealloc(Self { buffer: mut next_buffer, sender }: Self) {
+        while let Some(mut buffer) = next_buffer {
+            next_buffer = buffer.mut_next().take();
+            sender.blocking_send(buffer).expect("can't send buffer");
+        }
+    }
+
+    pub async fn dealloc(Self { buffer: mut next_buffer, sender }: Self) {
```
VadimYarovoy: Mb `release` is better name for this fn?

VadimYarovoy: Furthermore, `Self { buffer: mut next_buffer, sender}` in param looks strange - not obvious in code is it struct field or not

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+        while let Some(mut buffer) = next_buffer {
+            next_buffer = buffer.mut_next().take();
+            sender.send(buffer).await.expect("can't send blocking");
+        }
+    }
+}
+
+struct Allocator {
+    reciever: mpsc::Receiver<Buffer>,
```
VadimYarovoy: Types are already defined
```suggestion
    Receiver,
```

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+    sender: mpsc::Sender<Buffer>,
```
VadimYarovoy: Types are already defined

```suggestion
    Sender,
```

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+    size: usize,
+    count: usize,
+}
+
+impl Allocator {
+    pub async fn new(size: usize, count: usize) -> Self {
+        let (sender, reciever) = mpsc::channel::<Buffer>(count);
```
VadimYarovoy:
```suggestion
        let (sender, receiver) = mpsc::channel::<Buffer>(count);
```

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+        for _ in 0..count {
+            let buffer = Buffer::alloc(size);
+            sender.send(buffer).await.expect("cannot init buffers");
+        }
+
+        Self { sender, reciever, size, count }
+    }
+
+    pub async fn alloc(&mut self, mut size: usize) -> BufferChain {
+        assert!(size < self.size * self.count);
```
VadimYarovoy: why not eror?

```
if size > self.size * self.count {
    return Err("some msg or smothing else ....");
}
```

what are our actions in panic?

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        let mut chain = BufferChain::new(self.sender.clone());
+        while size != 0 {
+            let buffer = self.reciever.recv().await.expect("channel not to be closed");
+
+            chain.push(buffer);
+            size = size.wrapping_sub(self.size);
```
VadimYarovoy: why wrapping and not `saturating_sub`?

VadimYarovoy: to my mind it would be more intuitive if condition be next

```
let mut remaining = size;
        
while remaining > 0 {
    ...
    remaining = remaining.saturating_sub(self.size);
}

Ok(chain)
```

src/allocator.rs
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        chain
+    }
+}
+
+// Checks that: we can use public [`Buffer`] api.
+#[test]
```
VadimYarovoy: put tests in mod pls

VadimYarovoy: and add test for allocator overflow

src/allocator.rs
VadimYarovoy: I tried to do it here - https://github.com/RMamonts/nfs-mamont/pull/13

src/allocator.rs
IgorErin: Abort. Only abort (kill, actually). 

I's programmer error to request more than he allocates. He created this allocator, he knows with which parameters he created it. And it matches with global alloc behavior. 

I think we should treat this case as programmer error and kill process.

*On abort*

We should use abort to support WASM, for example. Unwinding is not portable

src/allocator.rs
IgorErin: Oh, vs code suggest incorrectly to me

src/allocator.rs
IgorErin: nice!

src/allocator/chain.rs
```diff
@@ -0,0 +1,58 @@
...
+            next_buffer = buffer.mut_next().take();
+            sender.blocking_send(buffer).expect("can't send blocking");
+        }
+    }
+
+    /// Deallocate in async manner.
+    pub async fn dealloc(Self { buffer: mut next_buffer, sender }: Self) {
+        while let Some(mut buffer) = next_buffer {
+            next_buffer = buffer.mut_next().take();
+            sender.send(buffer).await.expect("can't send async");
```
VadimYarovoy: Possible mem leak -  here is issue with full channel, we can do next

ignore errors to allow free it via Drop
```
let _ = sender.send(buffer).await;
```

IgorErin: Yea, i think like that

src/allocator/buffer.rs
```diff
@@ -0,0 +1,143 @@
...
+        // Safety: non zero layout size.
+        assert!(layout.size() != 0);
+        let ptr = unsafe { alloc::alloc_zeroed(layout) };
+        if ptr.is_null() {
+            alloc::handle_alloc_error(layout)
+        }
+
+        // Safety: non null, memory of size `size`, owning pointer
+        // Discassed: https://users.rust-lang.org/t/can-i-create-a-reference-to-a-custom-dst-from-raw-parts-on-stable/63261
+        assert!(!ptr.is_null());
```
VadimYarovoy: already checked with

```
if ptr.is_null() {
    alloc::handle_alloc_error(layout)
}
```

src/allocator/buffer.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Instrusive linked list heap allocated node layout.
+///
+/// Node allocated with Global allocator.
+#[repr(C)]
+struct BufferLayout {
+    // Pointer to the next entry of the intrusive linked list.
+    next: Option<Buffer>,
+    // Allocation layout.
+    layout: Layout,
+    // Actuall data.
```
VadimYarovoy:
```suggestion
    // Actual data.
```

VadimYarovoy:
@IgorErin update on main to allow checks to run

VadimYarovoy:
Here is update - https://github.com/RMamonts/nfs-mamont/pull/15

src/allocator/chain.rs
```diff
@@ -0,0 +1,105 @@
+use super::buffer::Buffer;
+use super::Sender;
```
IgorErin: Not Chain, just List

src/allocator/slice.rs
```diff
@@ -0,0 +1,41 @@
...
+impl Slice {
+    pub fn new(list: super::list::List, range: std::ops::Range<usize>) -> Self {
+        assert!(range.start <= list.len(), "cannot index list as slice from start");
+        assert!(range.end <= list.len(), "cannot index list as slice to end");
+
+        Self { list, range }
+    }
+}
+
+pub struct IterMut<'a> {
```
IgorErin, line 115: Add shared Iter

MikhailSvirukov (dismissed):

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
+//! Defines [`Slice`] --- list of buffers bounded by custome byte range.
+
+/// Represents bounded by custome range [`List`] of buffers.
```
IgorErin: No List

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Returns new [`Slice`] of specified buffers.
+    ///
+    /// # Parameters
+    ///
+    /// - `list` --- list of buffers.
+    /// - `range` --- range to which slice will allow access.
+    ///
+    /// # Panics
+    ///
+    /// This function will panics if called if length range bound greater then length of `list`.
```
IgorErin: buffers

IgorErin:
@artemiipatov @MikhailSvirukov look again, please

MikhailSvirukov (dismissed):

IgorErin:
@artemiipatov @VadimYarovoy @MikhailSvirukov review!

MikhailSvirukov (dismissed):

src/allocator/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    ///
+    /// # Parameters
+    ///
+    /// - `size` --- size of each buffer to allocate
+    /// - `count` --- number of buffers to allocate
+    pub fn new(size: NonZeroUsize, count: NonZeroUsize) -> Self {
+        let (sender, receiver) = mpsc::unbounded_channel::<Buffer>();
+
+        for _ in 0..count.get() {
+            sender.send(vec![0; size.get()].into_boxed_slice()).expect("to init Allocator");
```
artemiipatov: What does the expect message mean?

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
...
+            let len = result.len();
+            let start = self.range.start;
+            let end = self.range.end;
+
+            // make a progress
+            self.range.start = self.range.start.saturating_sub(len);
+            self.range.end = self.range.end.saturating_sub(len);
+
+            if len > start {
+                return Some(&result[self.range.start..end.min(len)]);
```
artemiipatov: Shouldn't we use saved value?
```suggestion
                return Some(&result[start..end.min(len)]);
```

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
...
+            let len = result.len();
+            let start = self.range.start;
+            let end = self.range.end;
+
+            // make a progress
+            self.range.start = self.range.start.saturating_sub(len);
+            self.range.end = self.range.end.saturating_sub(len);
+
+            if len > start {
+                return Some(&mut result[self.range.start..end.min(len)]);
```
artemiipatov: Same.

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub fn new(
+        buffers: Vec<Box<[u8]>>,
+        range: std::ops::Range<usize>,
+        sender: super::Sender<Box<[u8]>>,
+    ) -> Self {
+        let len = buffers.iter().map(|buffer| buffer.len()).sum();
+
+        assert!(range.start <= len, "cannot index list as slice from start");
+        assert!(range.end <= len, "cannot index list as slice to end");
+
```
artemiipatov, line 39:
```suggestion
        assert!(range.start <= range.end, "start should be less than or equal to end");
```

src/allocator/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    ///
+    /// - `size` --- size of returned slice.
+    ///
+    /// # Panic
+    ///
+    /// This method panics if size is greater then allocator capacity.
+    async fn alloc(&mut self, size: NonZeroUsize) -> Option<slice::Slice>;
+}
+
+pub struct Impl {
```
artemiipatov, line 35: I'm not quite sure, but probably we should use names like RecyleAllocator or CyclicAllocator

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+impl Slice {
+    /// Returns new [`Slice`] of specified buffers.
+    ///
+    /// # Parameters
+    ///
+    /// - `buffers` --- vec of buffers.
+    /// - `range` --- range to which slice will allow access.
+    ///
```
artemiipatov, line 18: Comment for `sender` parameter

artemiipatov (changes requested):

src/allocator/slice.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    pub fn iter(&self) -> Iter<'_> {
+        self.into_iter()
+    }
+
+    /// Deallocates all buffers i.e. send them via specified in the [`Self::new`] `sender`.
+    fn deallocate(&mut self) {
+        for mut buffer in self.buffers.drain(..) {
+            // No user data should exist after dealloc
+            buffer.iter_mut().for_each(|u| *u = 0);
```
artemiipatov: Memset.
```suggestion
            buffer.fill(0);
```

src/allocator/mod.rs
IgorErin: we expecte here to properly initialize allocator

artemiipatov:
```suggestion
            sender.send(vec![0; size.get()].into_boxed_slice()).expect("Allocator was not initialized properly");
```

src/allocator/slice.rs
IgorErin: Good!

IgorErin: Ironically,

```
impl<T: Copy> SpecFill<T> for [T] {
    fn spec_fill(&mut self, value: T) {
        for item in self.iter_mut() {
            *item = value;
        }
    }
}
```

src/allocator/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+use tokio::sync::mpsc;
+
+pub use slice::Slice;
+
+type Sender<T> = mpsc::UnboundedSender<T>;
+type Receiver<T> = mpsc::UnboundedReceiver<T>;
+type Buffer = Box<[u8]>;
+
+/// Allocates [`Slice`]'s.
+#[async_trait]
```
artemiipatov: What is the purpose of using `#[async_trait]` here? Are we going to use trait objects somewhere? Because if not, `#[async_trait]` is not necessary since 1.75v (if i am not mistaken).

IgorErin: Good point, should remove this

IgorErin: Good point, should remove this

IgorErin: Good point, should remove this

src/allocator/mod.rs
IgorErin: Not to contain work allocator in public entity in module `allocator`. We should use modules for such a naming.

Ok, we throw away postifx at that point, then `Recycle` or `Cyclic` allocator remains. Every allocator is Recycle and Cyclic since it allocates and then deallocates. Why not just allocator::Impl? It will be invisible to users after allocator::Service implementation

IgorErin:
@artemiipatov @MikhailSvirukov look again, please

artemiipatov (approved):

MikhailSvirukov (approved):

PR #16, opened 2025-10-18
Xdr deserializer

MikhailSvirukov:
- add functions to parse primitive types
- change vec/strings that have limits by protocols to specific types to ease later parsing


src/mount.rs
```diff
@@ -1,19 +1,21 @@
...
 #[allow(dead_code)]
 const MOUNT_PROGRAM: u32 = 100005;
 #[allow(dead_code)]
 const MOUNT_VERSION: u32 = 3;
 #[allow(dead_code)]
-const MNTPATHLEN: u32 = 1024;
+const MNTPATHLEN: usize = 1024;
 #[allow(dead_code)]
-const MNTNAMLEN: u32 = 255;
+const MNTNAMLEN: usize = 255;
```
IgorErin: Why usize?

src/parser/mod.rs
```diff
@@ -0,0 +1,13 @@
+pub mod to_parse;
+
+#[derive(Debug)]
+#[allow(unused)]
+pub enum ParserError {
+    StringTooLong,
+    ReadError,
```
IgorErin: IOError?

src/parser/mod.rs
```diff
@@ -0,0 +1,13 @@
+pub mod to_parse;
+
+#[derive(Debug)]
+#[allow(unused)]
+pub enum ParserError {
```
IgorErin: Just error, ot already in parse module

src/parser/to_parse.rs
```diff
@@ -0,0 +1,378 @@
+use std::io::Read;
+use std::mem::MaybeUninit;
+
+use super::ParserError;
+use byteorder::{BigEndian, ReadBytesExt};
```
IgorErin: Import order

IgorErin (changes requested):
Do we need trait ToParse?

MikhailSvirukov:
> Do we need trait ToParse?

I think we should move some of logic from Parser, that would do argument structures. This parsing should know nothing about allocator or socket. But while parsing of actual argument of procedure might want to know that (especially when receiving write request)

src/parser/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod to_parse;
+
+#[derive(Debug)]
+#[allow(unused)]
+pub enum Error {
+    StringTooLong,
+    VecTooLong,
+    IO,
```
IgorErin: Wrap std::io::Error inside this variant

src/parser/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+    EnumDiscMismatch,
+    ArrayConvert,
+    IncorrectString,
```
IgorErin: String has it's own errors for parsing, i think. Can we wrap them here too?

IgorErin:
> > Do we need trait ToParse?
> 
> I think we should move some of logic from Parser, that would do argument structures. This parsing should know nothing about allocator or socket. But while parsing of actual argument of procedure might want to know that (especially when receiving write request)

I see, but for what trait here? Mb just functions? I give you example in review

src/parser/to_parse.rs
```diff
@@ -0,0 +1,391 @@
...
+            vec.push(T::parse(src)?);
+        }
+        Ok(vec)
+    }
+}
+#[allow(dead_code)]
+pub struct VecWithMaxLen<const N: u32>(Vec<u8>);
+impl<const N: u32> ToParse for VecWithMaxLen<N> {
+    fn parse<R: Read>(src: &mut R) -> Result<Self, Error> {
+        let size = u32::parse(src)?;
```
IgorErin: fn u32_as_usize

src/parser/to_parse.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+pub fn read_padding<R: Read>(src: &mut R, n: usize) -> Result<(), Error> {
+    let mut buf = [0u8; ALIGNMENT];
+    let padding = (ALIGNMENT - n % ALIGNMENT) % ALIGNMENT;
+    src.read_exact(&mut buf[..padding]).map_err(|_| Error::IncorrectPadding)
+}
+
+impl ToParse for u32 {
+    fn parse<R: Read>(src: &mut R) -> Result<Self, Error> {
```
IgorErin: why not just

```
fn u32(srd: &mut R) -> Result<Self, Error> {
     src.read_u32::<BigEndian>().map_err(|_| Error::IO)
}
```

for what trait?

Here we can return exactly std::io::Result actually

src/parser/to_parse.rs
```diff
@@ -0,0 +1,391 @@
...
+        let discr = parse::<u32, R>(src)?;
+        match discr {
+            0 => Ok(false),
+            1 => Ok(true),
+            _ => Err(Error::EnumDiscMismatch),
+        }
+    }
+}
+
+impl<T: ToParse> ToParse for Option<T> {
```
IgorErin: For option, i see, ok. I should thin for a moment

IgorErin (changes requested):

src/parser/to_parse.rs
MikhailSvirukov: Yes, i also though about that. But what should we do with types like, for example `sattr3`? I don't think that we should explicitly construct them in Parser, because it won't look pretty and also it's not very expandable. 

I see it like this: 
```
Parser (part that reads from socket and manages memory) 
                              /\
                              ||
                              ||
                              \/ 
separate functions that parse stream of bytes into actual structure, representing proc arguments (probably wrap them in our own types? Maybe separate module?) 
                              /\
                              ||
                              ||
                              \/ 
ToParse trait to parse fields of structures
```

So, answering your question, trait is used to have universal way to parse fields of structures, because they can be both primitive types of structures (and enums)

IgorErin: Trait not expandable since it fixes interface. For example: 
```
fn option<T>(r: &mut impl Read, parse_inner: FnOnce(&mut impl Read) -> T) -> std::io::Result<Option<T>> {
     if read_u32(r)? == 1 {
         return parse_inner(r)?
     } 
    
     Ok(None)
}
```

here i can parse everything inside option with variable number of arguments (variable number of allocators) 
but `ToParse` limit me to only one `&mut iml Read` and unified parsing VecWithFixedSize with u32 which are different to my thinking

src/parser/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+    IncorrectFilehandle,
```
IgorErin: What mean IncorrectFilehandle?

src/parser/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod to_parse;
+
+#[derive(Debug)]
+#[allow(unused)]
+pub enum Error {
+    StringTooLong,
+    VecTooLong,
```
IgorErin: What mean Vec too long?

src/parser/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod to_parse;
+
+#[derive(Debug)]
+#[allow(unused)]
+pub enum Error {
+    StringTooLong,
```
IgorErin: What mean string too long?

src/parser/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+    VecTooLong,
+    IO,
+    EnumDiscMismatch,
+    ArrayConvert,
```
IgorErin: What mean array convert? Mb add comments?

src/parser/primitive.rs
```diff
@@ -0,0 +1,325 @@
...
+    let mut vec = Vec::with_capacity(size);
+    for _ in 0..size {
+        vec.push(cont(src)?);
+    }
+    read_padding(src, size_of::<T>() * size)?;
+    Ok(vec)
+}
+
+#[allow(dead_code)]
+pub fn parse_vec_max_size<const N: usize, T>(
```
IgorErin: Not as const, just parameter.

For every const generic we generate new function which leads as to code bloat and reduced performanc

src/parser/nfsv3.rs
```diff
@@ -0,0 +1,365 @@
...
+        4 => Ok(mknoddata3::NF3CHR(parse_devicedata3(src)?)),
+        5 => Ok(mknoddata3::NF3LNK),
+        6 => Ok(mknoddata3::NF3SOCK(parse_sattr3(src)?)),
+        7 => Ok(mknoddata3::NF3FIFO(parse_sattr3(src)?)),
+        _ => Err(Error::EnumDiscMismatch),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
IgorErin: Mb in separate module?

src/rpc.rs
```diff
@@ -2,6 +2,8 @@
 
 #[allow(dead_code)]
 const RPC_VERSION: u32 = 2;
+#[allow(dead_code)]
+const MAX_AUTH_OPAQUE: u32 = 400;
```
IgorErin: But what a MAX? Len, size?

IgorErin (changes requested):
Separate mod for tests?

src/parser/nfsv3.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::parser::primitive::{
+    parse_array, parse_option, parse_string_max_len, parse_u32, parse_u64, parse_u8,
+    parse_vec_max_size,
+};
+use crate::parser::Error;
+
+#[allow(dead_code)]
+const MAX_FILENAME: usize = 255;
+#[allow(dead_code)]
+pub const MAX_FILEHANDLE: usize = 255;
```
IgorErin:
```suggestion
pub const MAX_FILEHANDLE: usize = 8;
```

src/nfsv3.rs
```diff
@@ -9,9 +11,13 @@ const NFS3_FHSIZE: u32 = 64;
 #[allow(dead_code)]
 const NFS3_COOKIEVERFSIZE: u32 = 8;
 #[allow(dead_code)]
-const NFS3_CREATEVERFSIZE: u32 = 8;
+pub const NFS3_CREATEVERFSIZE: usize = 8;
 #[allow(dead_code)]
 const NFS3_WRITEVERFSIZE: u32 = 8;
+#[allow(dead_code)]
+const MAX_FILENAME_SIZE: u32 = 255;
```
IgorErin, line 18: 8?

tests/parser.rs
```diff
@@ -0,0 +1,453 @@
...
+    use nfs_mamont::nfsv3::{ftype3, mknoddata3};
+    use nfs_mamont::parser::nfsv3::parse_specdata3;
+    use nfs_mamont::parser::nfsv3::{
+        parse_createhow3, parse_devicedata3, parse_diropargs3, parse_mknoddata3, parse_nfs_fh3,
+        parse_nfstime, parse_sattr3, parse_set_atime, parse_set_mtime, parse_symlinkdata3,
+        MAX_FILEHANDLE,
+    };
+    use nfs_mamont::parser::primitive::parse_c_enum;
+    use nfs_mamont::parser::Error;
+    use std::io::Cursor;
```
IgorErin: import order

src/parser/primitive.rs
```diff
@@ -0,0 +1,128 @@
...
+    String::from_utf8(vec).map_err(Error::IncorrectString)
+}
+
+#[allow(dead_code)]
+pub fn parse_c_enum<T: FromPrimitive>(src: &mut dyn Read) -> Result<T, Error> {
+    let val = FromPrimitive::from_u32(parse_u32(src)?);
+    match val {
+        Some(res) => Ok(res),
+        None => Err(Error::EnumDiscMismatch),
+    }
```
IgorErin:
```suggestion
     FromPrimitive::from_u32(parse_u32(src)?).ok_or(Error::EnumDiscMismatch)

```

src/parser/primitive.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[allow(dead_code)]
+fn parse_u32_as_usize(src: &mut dyn Read) -> Result<usize, Error> {
+    match parse_u32(src)?.to_usize() {
+        None => Err(Error::ImpossibleTypeCast),
+        Some(n) => Ok(n),
+    }
```
IgorErin: use .ok_or for Option -> Result conversion

src/parser/primitive.rs
```diff
@@ -0,0 +1,128 @@
...
+        true => Ok(Some(cont(src)?)),
+        false => Ok(None),
+    }
+}
+
+#[allow(dead_code)]
+pub fn parse_array<const N: usize, T>(
+    src: &mut dyn Read,
+    cont: impl Fn(&mut dyn Read) -> Result<T, Error>,
+) -> Result<[T; N], Error> {
```
IgorErin:
```suggestion
) -> Result<Box<[T; N]>, Error> {
```

IgorErin: don't pass array on stack

src/parser/mod.rs
```diff
@@ -0,0 +1,16 @@
+use std::io;
+use std::string::FromUtf8Error;
+
+pub mod nfsv3;
+pub mod primitive;
+
```
IgorErin, line 8:
```suggestion
type Result<T> = std::result::Result<T, Error>;
```

src/parser/nfsv3.rs
```diff
@@ -0,0 +1,112 @@
...
+
+#[allow(dead_code)]
+pub fn parse_devicedata3(src: &mut dyn Read) -> Result<devicedata3, Error> {
+    Ok(devicedata3 { dev_attributes: parse_sattr3(src)?, spec: parse_specdata3(src)? })
+}
+
+#[allow(dead_code)]
+pub fn parse_mknoddata3(src: &mut dyn Read) -> Result<mknoddata3, Error> {
+    let disc = parse_u32(src)?;
+    match disc {
```
IgorErin:
```suggestion
    match parse_u32(src)? {
```

IgorErin: small expression unused name, mb inline?

src/parser/nfsv3.rs
```diff
@@ -0,0 +1,112 @@
...
+        1 => Ok(set_atime::SET_TO_SERVER_TIME),
+        2 => Ok(set_atime::SET_TO_CLIENT_TIME(parse_nfstime(src)?)),
+        _ => Err(Error::EnumDiscMismatch),
+    }
+}
+
+#[allow(dead_code)]
+pub fn parse_set_mtime(src: &mut dyn Read) -> Result<set_mtime, Error> {
+    let disc = parse_u32(src)?;
+    match disc {
```
IgorErin:
```suggestion
    match parse_u32(src)? {
```

IgorErin (changes requested):
Small changes?

src/parser/primitive.rs
MikhailSvirukov: You sure we always need it in heap?

IgorErin:
and update description, please

IgorErin:
mb with no `parse_` prefix

src/parser/mod.rs
```diff
@@ -0,0 +1,19 @@
+use std::io;
+use std::string::FromUtf8Error;
+
+pub mod nfsv3;
+pub mod primitive;
```
IgorErin:
```suggestion
pub mod nfsv3;
pub mod primitive;

use std::io;
use std::string::FromUtf8Error;
```

src/parser/nfsv3.rs
```diff
@@ -0,0 +1,110 @@
...
+    array, option, string_max_len, to_u32, to_u64, to_u8, u32_as_usize,
+};
+use crate::parser::{Error, Result};
+
+#[allow(dead_code)]
+const MAX_FILENAME: usize = 255;
+#[allow(dead_code)]
+pub const MAX_FILEHANDLE: usize = 8;
+#[allow(dead_code)]
+const MAX_FILEPATH: usize = 255;
```
IgorErin: we have max file path set to 1024 in vfs.rs. what do you think about it?

tests/parser.rs
```diff
@@ -0,0 +1,450 @@
+mod nfsv3 {
```
IgorErin: mb separate files?

IgorErin: And we can place tests inside src/parser directory, if you prefer. 

no need in integration tests (that lives outside of lib crate)

src/parser/nfsv3.rs
```diff
@@ -0,0 +1,110 @@
...
+#[allow(dead_code)]
+pub fn diropargs3(src: &mut dyn Read) -> Result<diropargs3> {
+    Ok(diropargs3 { dir: nfs_fh3(src)?, name: string_max_len(src, MAX_FILEPATH)? })
+}
+
+#[allow(dead_code)]
+pub fn createhow3(src: &mut dyn Read) -> Result<createhow3> {
+    match to_u32(src)? {
+        0 => Ok(createhow3::UNCHECKED(sattr3(src)?)),
+        1 => Ok(createhow3::UNCHECKED(sattr3(src)?)),
```
IgorErin:
```suggestion
        1 => Ok(createhow3::GUARDED(sattr3(src)?)),
```

?

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+use byteorder::{BigEndian, ReadBytesExt};
+use num_traits::{FromPrimitive, ToPrimitive};
+
+use super::{Error, Result};
+
+#[allow(dead_code)]
+pub const ALIGNMENT: usize = 4;
+
+#[allow(dead_code)]
+fn read_padding(src: &mut dyn Read, n: usize) -> Result<()> {
```
IgorErin:
```suggestion
fn padding(src: &mut dyn Read, n: usize) -> Result<()> {
```

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+    let mut buf = [0u8; ALIGNMENT];
+    let padding = (ALIGNMENT - n % ALIGNMENT) % ALIGNMENT;
+    src.read_exact(&mut buf[..padding]).map_err(|_| Error::IncorrectPadding)
+}
+
+#[allow(dead_code)]
+pub fn to_u8(src: &mut dyn Read) -> Result<u8> {
```
IgorErin:
```suggestion
pub fn u8(src: &mut dyn Read) -> Result<u8> {
```

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+    src.read_u8().map_err(Error::IO)
+}
+
+#[allow(dead_code)]
+pub fn to_u32(src: &mut dyn Read) -> Result<u32> {
```
IgorErin:
```suggestion
pub fn u32(src: &mut dyn Read) -> Result<u32> {
```

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+    src.read_u32::<BigEndian>().map_err(Error::IO)
+}
+
+#[allow(dead_code)]
+pub fn to_u64(src: &mut dyn Read) -> Result<u64> {
```
IgorErin:
```suggestion
pub fn u64(src: &mut dyn Read) -> Result<u64> {
```

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+    src.read_u64::<BigEndian>().map_err(Error::IO)
+}
+
+#[allow(dead_code)]
+pub fn to_bool(src: &mut dyn Read) -> Result<bool> {
```
IgorErin:
```suggestion
pub fn bool(src: &mut dyn Read) -> Result<bool> {
```

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+    let mut vec = Vec::with_capacity(size);
+    for _ in 0..size {
+        vec.push(cont(src)?);
+    }
+    read_padding(src, size_of::<T>() * size)?;
+    Ok(vec)
+}
+
+#[allow(dead_code)]
+pub fn string_max_len(src: &mut dyn Read, max_size: usize) -> Result<String> {
```
IgorErin:
```suggestion
pub fn string_max_size(src: &mut dyn Read, max_size: usize) -> Result<String> {
```
?

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[allow(dead_code)]
+pub fn string(src: &mut dyn Read) -> Result<String> {
+    let vec = vector(src, |s| to_u8(s))?;
+    String::from_utf8(vec).map_err(Error::IncorrectString)
+}
+
+#[allow(dead_code)]
+pub fn c_enum<T: FromPrimitive>(src: &mut dyn Read) -> Result<T> {
```
IgorErin:
```suggestion
pub fn union<T: FromPrimitive>(src: &mut dyn Read) -> Result<T> {
```

Just union?

IgorErin (changes requested):
Seems like typo in varian parsing, and small changes

tests/parser.rs
MikhailSvirukov: If it a normal practice, than ok. It just looks a bit strange for me, that we have separate module with tests inside one that have source code (i've never seen such)

src/parser/primitive.rs
MikhailSvirukov: Name conflicts with type names

src/parser/nfsv3.rs
MikhailSvirukov: I understand, that it is a little concern, in case of constants, but what is connection between modules? Probably we shouldn't redefine constants (basically i do it here). Then, where should that one constant be? It doesn't look too bad, but someday we might get confused, since there are quite many constants

src/parser/primitive.rs
MikhailSvirukov: Why union? I meant c-style enum

IgorErin: Oh, I see, then mb `variant`? I understand that you mean c-style enum. I can't understand why it `c-style`

src/parser/primitive.rs
IgorErin: Where? How function name can conflict with type name?

IgorErin: image

src/parser/nfsv3.rs
IgorErin: I disagree with you. Here constants that defines maximum length for string to be parsed. Constant in `vfs.rs` module defined maximum name length for passed into `Vfs` trait. And in `nfs3.rs` (files that were written directly from RFC) constant defines some unneeded trash as i see it today but let's remove it later.

Then we got two interfaces here: 
- NFS server client interface --- constant in parsing
- NFS server user interface --- constant in `vfs.rs`

And next statement should stay true: `vfs::MAX_FILEPATH <= parser::MAX_FILEPATH`

In this comment I just point your attention to that fact and ask you about change this constant to 1024 units

src/nfsv3.rs
IgorErin: One of them usize, remaining -- u32. Why?

src/parser/nfsv3.rs
MikhailSvirukov: Ok. Then we should add some additional checks about path size in vfs and probably write somewhere in docs about inequality

src/parser/primitive.rs
MikhailSvirukov: I get some strange [error](https://doc.rust-lang.org/error_codes/E0747.html) after renaming in my IDE, though code compiles

MikhailSvirukov:
@artemiipatov @IgorErin 
Can you, please, look at this one?)

src/parser/tests/nfsv3.rs
```diff
@@ -0,0 +1,249 @@
+#![cfg(test)]
```
IgorErin: For what it here?

src/parser/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+mod tests;
```
IgorErin, line 4: Here should be #[cfg(test)]

IgorErin (dismissed):

MikhailSvirukov:
@artemiipatov @VadimYarovoy

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+    for elem in buf.iter_mut() {
+        elem.write(cont(src)?);
+    }
+    padding(src, N * size_of::<T>())?;
+    Ok(unsafe { buf.as_ptr().cast::<[T; N]>().read() })
+}
+
+#[allow(dead_code)]
+pub fn vector<T>(src: &mut dyn Read, cont: impl Fn(&mut dyn Read) -> Result<T>) -> Result<Vec<T>> {
+    let size = u32_as_usize(src)?;
```
artemiipatov, line 69: Should we add a check for the incoming size? Or do you plan to add it in the future PR dedicated to validation? Otherwise i think, this function (and `string`) should not be pub.

src/parser/primitive.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[allow(dead_code)]
+pub fn array<const N: usize, T>(
+    src: &mut dyn Read,
+    cont: impl Fn(&mut dyn Read) -> Result<T>,
+) -> Result<[T; N]> {
+    let mut buf: [MaybeUninit<T>; N] = [const { MaybeUninit::uninit() }; N];
+    for elem in buf.iter_mut() {
+        elem.write(cont(src)?);
```
artemiipatov, line 61: I am not quite sure, but shouldn't we drop already initialized MaybeUninit values explictily to prevent memory leak in case of error from cont(src)?

artemiipatov (changes requested):
If I understood the intention correctly, we are gonna parse incoming data on the fly passing Read channel to our primitives when needed. Then, i guess, they should contain checks to prevent ddos. Or are we gonna add them in the future PR, dedicated to validation?

src/parser/primitive.rs
MikhailSvirukov: That's definitely not something we should do here. This function simply parse as if there were no server. 
I do understand necessity of such checks, but it would be done in, probably, separate actor "Validator". Parser, in my opinion, should simply receive some bytes and convert them in some convenient representation, checking correctness in terms of RFC (!)

src/parser/primitive.rs
MikhailSvirukov: I think we could limit that to Copy types. Than it would be okay (since we don't call drop). 
Overall i think it might be a good idea to limit bounds of `T`. This sound simple + we don't actually use this other than with numeric types

MikhailSvirukov: @IgorErin, what do you think?

artemiipatov (approved):

IgorErin (approved):

src/parser/primitive.rs
IgorErin: @MikhailSvirukov i think we ok here for now

PR #17, opened 2025-10-20
vfs: no null procedure

IgorErin:


src/vfs.rs
```diff
@@ -371,31 +309,29 @@ pub struct CommitResult {
...
+    async fn get_attr(&self, handle: &FileHandle) -> io::Result<FileAttr>;
 
     /// Procedure 2: SETATTR – mutate file attributes with optional guard.
     async fn set_attr(
         &self,
         handle: &FileHandle,
         attr: SetAttr,
         guard: SetAttrGuard,
-    ) -> VfsResult<WccData>;
+    ) -> io::Result<WccData>;
```
MikhailSvirukov: You sure, it's a good idea? How are we going to map errors to protocol errors, when we have to send reply to client with error code, for example `nfsstat3`?

MikhailSvirukov (approved):

VadimYarovoy (approved):

PR #19, opened 2025-10-26
feat: file handle const size

artemiipatov:
Accroding to RFC 1813 fhandle size is variable and cannot be more than 64 bytes. But in practice, it is not possible to overflow 8 bytes, and it is easier to work with contant size. So we decided to set it to 8 bytes.

### DIFF
--- a/src/mount.rs
+++ b/src/mount.rs
@@ -11,7 +11,7 @@ const MNTNAMLEN: u32 = 255;
 #[allow(dead_code)]
 const FHSIZE3: u32 = 64;
 
-type fhandle3 = Vec<u8>;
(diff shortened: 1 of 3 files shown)


MikhailSvirukov (dismissed):

IgorErin:
@artemiipatov add to description which size and why

IgorErin (changes requested):
Why this size? Mb more words in description

IgorErin:
@artemiipatov `MAX_FILE_HANDLE_LEN` in vfs.rs also should be set?

IgorErin (approved):

MikhailSvirukov (approved):

PR #20, opened 2025-10-27
Implement ShadowFS: A filesystem mirroring the host filesystem

### LINKED ISSUE #12: Implement VFS trait
we need demo example with implemented functions

VadimYarovoy:
Resolves #12

### DIFF
--- a/examples/shadow_fs/fs/mod.rs
+++ b/examples/shadow_fs/fs/mod.rs
@@ -1,0 +1,596 @@
+use std::cmp;
+use std::ffi::{OsStr, OsString};
+use std::io;
+use std::os::unix::fs::{MetadataExt, PermissionsExt};
+use std::path::Path;
+
+use tokio::fs::{self, File, OpenOptions};
+use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt};
+use tokio::task;
+
+use nfs_mamont::vfs;
+
+mod shadow;
+mod state;
+mod utils;
+
+pub use shadow::ShadowFS;
+
+use utils::{
+    apply_setattr, ensure_supported_attr, file_name_string, join_child, map_io_error,
+    ENTRY_ESTIMATE_BYTES,
+};
+
+#[async_trait::async_trait]
+impl vfs::Vfs for ShadowFS {
+    async fn get_attr(&self, handle: &vfs::FileHandle) -> vfs::VfsResult<vfs::FileAttr> {
+        let id = Self::decode_handle(handle)?;
+        let meta = self.metadata_for_id(id).await?;
+        Ok(Self::attr_from_meta(id, &meta))
+    }
+
+    async fn set_attr(
+        &self,
+        handle: &vfs::FileHandle,
+        attr: vfs::SetAttr,
+        guard: vfs::SetAttrGuard,
+    ) -> vfs::VfsResult<vfs::WccData> {
+        let id = Self::decode_handle(handle)?;
+        let rel = self.rel_path_from_id(id).await?;
+        let abs = self.full_path(&rel);
+        let before_meta = self.metadata_for_rel(&rel).await?;
+        let before_attr = Self::attr_from_meta(id, &before_meta);
+
+        if let vfs::SetAttrGuard::Check { ctime } = guard {
+            if before_attr.ctime != ctime {
+                return Err(vfs::NfsError::NotSync);
+            }
+        }
+
+        apply_setattr(&abs, &attr).await?;
+
+        let after_meta = self.metadata_for_rel(&rel).await?;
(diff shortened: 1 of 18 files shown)


examples/shadow_fs/fs/mod.rs
```diff
@@ -0,0 +1,596 @@
...
+use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt};
+use tokio::task;
+
+use nfs_mamont::vfs;
+
+mod shadow;
+mod state;
+mod utils;
+
+pub use shadow::ShadowFS;
```
IgorErin, line 17: Why shadow?

examples/shadow_fs/fs/mod.rs
```diff
@@ -0,0 +1,596 @@
+use std::cmp;
```
IgorErin, line 1: Let's use binary! For what this example is?

IgorErin (changes requested):
This one before: https://github.com/RMamonts/nfs-mamont/pull/21

Please, doesn't generate code by AI any more. This is unreadable and error prone

examples/shadow_fs/fs/mod.rs
VadimYarovoy: Binary will be is src/, I think we don't need it in src

examples/shadow_fs/fs/mod.rs
VadimYarovoy: Name from my head, use it not to name it mirrorfs, We can use any name

VadimYarovoy:
@IgorErin yea it is partly generated, but it still will be a lot of code without generated one.

My goal is to write tests for it and remove or fix if something will be incorrect.

For now we need something that works as fast as possible.

I can add Vfs procedure by procedure but this will take a lot of time

VadimYarovoy:
@IgorErin will try look at you pr #21 today and change this one

examples/shadow_fs/fs/mod.rs
IgorErin: why&

examples/shadow_fs/fs/mod.rs
IgorErin: why not just mamontfs?

examples/shadow_fs/fs/mod.rs
IgorErin: Our target for now --- NFS3 compatible standalone implementation, why create another implementation in examples?

IgorErin:
@VadimYarovoy Ok, i just mention some redundant moments in vfs.rs by looking into you current PR and (as i thought) try to fix it ASAP in mentioned PR to allow to simplify this implementation

examples/shadow_fs/fs/utils.rs
```diff
@@ -0,0 +1,149 @@
...
+    match err.kind() {
+        NotFound => vfs::NfsError::NoEnt,
+        PermissionDenied => vfs::NfsError::Access,
+        AlreadyExists => vfs::NfsError::Exist,
+        InvalidInput | InvalidData => vfs::NfsError::Inval,
+        NotADirectory => vfs::NfsError::NotDir,
+        IsADirectory => vfs::NfsError::IsDir,
+        ReadOnlyFilesystem => vfs::NfsError::RoFs,
+        StorageFull | OutOfMemory => vfs::NfsError::NoSpc,
+        _ => vfs::NfsError::Io,
```
artemiipatov, line 26: Could you explain, what is the reason of duplicating nfsstat3? Why cannot we just use nfsv3::nfsstat3 instead of vfs::NfsError?

examples/shadow_fs/fs/mod.rs
```diff
@@ -0,0 +1,596 @@
...
+        offset: u64,
+        data: &[u8],
+        mode: vfs::WriteMode,
+    ) -> vfs::VfsResult<vfs::WriteResult> {
+        let id = Self::decode_handle(handle)?;
+        let rel = self.rel_path_from_id(id).await?;
+        let abs = self.full_path(&rel);
+        let mut file = OpenOptions::new().write(true).open(&abs).await.map_err(map_io_error)?;
+        file.seek(io::SeekFrom::Start(offset)).await.map_err(map_io_error)?;
+        file.write_all(data).await.map_err(map_io_error)?;
```
artemiipatov, line 181: Why are there no limits on write and read requests sizes? Or do we plan to add them in next PRs?

examples/shadow_fs/fs/utils.rs
IgorErin: @artemiipatov because this is generated code...

VadimYarovoy:
useless

Issue #3, opened 2025-10-06
Closing tcp connection


MikhailSvirukov:
There is no description given, whether we should close socket or not , after producing error on rpc request. So, probably, we should define that behavior, since it is not protocol-related

MikhailSvirukov:
I see two options:

- simply close connection, after sending reply, and drop every related task
- setup socket with [keepalive](https://datatracker.ietf.org/doc/html/rfc1122#page-101) using [socket2](https://docs.rs/socket2/latest/socket2/) crate 

@IgorErin @VadimYarovoy what do you think?

VadimYarovoy:
@MikhailSvirukov @IgorErin I prefer second option

> setup socket with [keepalive](https://datatracker.ietf.org/doc/html/rfc1122#page-101) using [socket2](https://docs.rs/socket2/latest/socket2/) crate

for next reasons:
- Reconnection is expensive (handshake, auth, and so on)
- Many NFS errors (NOENT, ACCES, PERM, ..) are normal situations and not a reason to close the connection

but next list is for cases when we should terminate connection
- I/O errors (but here probably will be client connection abort)
- Critical internal server errors (panic, OOM, ...)
- RPC message parsing errors (incorrect client o our bug)

PR #25, opened 2025-12-01
Procedure parser addition

MikhailSvirukov:
- add main structure, that will handle parsing of byte stream
- add several nested objects to separate counters logic

Previously, we limited all RFC defined types, that has no upper bounds of size. That made all our arguments limited (except for one - write); that means, that we can process them with single buffer. Handling writes is a separate case, that will use same model for parsing procedure headers, but `Slice`  with async alloc, introduced in #11. So basically, we are going to reuse limited buffer, but also use Slice in case of `write` to avoid heavy coping


MikhailSvirukov:
Overall, i have come to two major conclusions:

1. It is not possible to implement our parser with making read from socket only when there is no data in buffer (as i initially tried). Main problem for that is the way we parse arrays/vectors - through passing [closure](https://github.com/RMamonts/nfs-mamont/blob/4d05f6cbc8e9d05ea6e15211304f9b05bc4676af/src/parser/primitive.rs#L55). That breaks all logic of self-managing cursor position inside buffer. So, i do implement this with 2 read calls - one for header of RPC message (fixed part of it) and one for the rest of the message.
2. Our model with fixed inner buffer and async allocator cannot normally manage parsing message, if previous message ended up with error (initial discussion - #3).  If we need to read remaining bytes from socket, we do have 2 options: sequentially read in fixed buffer (breaks cursor logic of fixed-size buffer) or use asunc allocator, which conflicts with purpose of allocator. 

So, i suggest we do switch to several read calls (which is not that bad, i suppose, ganesha does the same; also it really simplifies cursor managing) and dropping tcp connection after sending error reply

MikhailSvirukov:
This version is not relevant in favor of #26

PR #24, opened 2025-12-01
NFSv3: procedures parsing

MikhailSvirukov:
- define structures, representing arguments for procedures
- create  nfsv3 procedures parser, using primitive parsers

basically, this code parses types, defined in RFC (`src/nfsv3 `mod) and converts them in our inner representation

PR #21, opened 2025-10-28
vfs: more correct and more granular

IgorErin:
Seems like many `vfs` structures generated by AI.

Rewrite it, add more comments and split NFS function 
into separate traits with promise result forwarding

Resolves: #12

PR #26, opened 2025-12-10
Parser improved version

MikhailSvirukov:
First variant appeared in #25  
This version present an improved one, that resolves problems with previous one:
* rearrange logic to support cursor invariant in case of fails: that includes explicit saving before any calls
* refactor functions of array/vector parsing: change them to explicit u8, since there is no use case for polymorphism 
* implement separate logic to support parsing after non-critical errors


src/parser/read_buffer.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.total_bytes += from_inner;
+        let from_socket = n - from_inner;
+        if from_socket == 0 {
+            return Ok(());
+        }
+
+        let mut src = (&mut self.socket).take(from_socket as u64);
+
+        let mut dest = tokio::io::sink();
+        let actual = tokio::io::copy(&mut src, &mut dest).await?;
```
MikhailSvirukov: I am not sure about this function.
`copy()` creates on every call underlying buffer with default size of 8kbyte, which is not good, since we are trying to handle allocations gently.
Apparently, there is alternative with `AsyncBufRead`, but i think it is not a good idea to restrict socket to it, because we already have one buffering logic, and one more buffering would result in just more copying. But, on the other hand it does provide 'AsyncSeek' implementation (which we might use)
So, apparently, the simplest (and easiest) solution - create one more buffer to read bytes. 
@IgorErin, what do you think?

VadimYarovoy:
Do we need to merge this #24 before?

MikhailSvirukov:
> Do we need to merge this #24 before?

Yes

VadimYarovoy:
@MikhailSvirukov ready for review?

VadimYarovoy:
@MikhailSvirukov here are conflicts with main

IgorErin:
@MikhailSvirukov check test structure and format that i used in #21

src/parser/tests/parser_struct.rs
```diff
@@ -0,0 +1,147 @@
...
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x08, 0x01, 0x02, 0x03, 0x04, 0x05,
+        0x06, 0x07, 0x08, 0x80, 0x00, 0x00, 0x30, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x02, 0x00, 0x01, 0x86, 0xA3, 0x00, 0x00, 0x00, 0x03, 0x00, 0x00, 0x00,
+        0x12, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x08, 0x01, 0x02,
+        0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
+    ];
+    let socket = MockSocket::new(buf.as_slice());
+    let alloc = MockAllocator::new(0);
+    let mut parser = RpcParser::new(socket, alloc, 0x50);
+    let result1 = parser.parse_message().await;
```
VadimYarovoy: Use just `result` if don't check `result_i` in futher code

src/parser/nfsv3/procedures.rs
```diff
@@ -195,38 +199,66 @@ pub fn read(src: &mut dyn Read) -> Result<ReadArgs> {
...
-        object: FileHandle(nfs_fh3(src)?.data),
-        offset: u64(src)?,
-        count: u32(src)?,
-        mode: write_mode(src)?,
-        data: {
-            read_in_slice(src, &mut slice)?;
-            slice
-        },
-    })
+pub fn write(src: &mut dyn Read) -> Result<(FileHandle, u64, u32, WriteMode, usize)> {
```
VadimYarovoy: What are the meanings of `u64`, `u32` and `usize` in `Result<(FileHandle, u64, u32, WriteMode, usize)>`?

Possible solutions can be some structs with named fields or documentation above this fn

src/nfsv3.rs
```diff
@@ -3,9 +3,9 @@
 use num_derive::FromPrimitive;
 
 #[allow(dead_code)]
-const NFS_PROGRAM: u32 = 100003;
+pub(crate) const NFS_PROGRAM: u32 = 100003;
 #[allow(dead_code)]
-const NFS_VERSION: u32 = 3;
+pub(crate) const NFS_VERSION: u32 = 3;
```
VadimYarovoy, line 8: Why `pub(crate)` and `#[allow(dead_code)]`? Moreover try not to use `pub(crate)`

VadimYarovoy: UPD: here  `#[allow(dead_code)]` is useless, `pub(crate)` too, because in tests you do

```
use crate::nfsv3::{NFS_PROGRAM, NFS_VERSION};
```

src/parser/mount.rs
```diff
@@ -0,0 +1,35 @@
+use std::io::Read;
+
+use super::Result;
+use crate::parser::primitive::string_max_size;
+use crate::vfs::{FsPath, MAX_PATH_LEN};
+
+#[allow(dead_code)]
+enum MountStat {
```
VadimYarovoy, line 10: why this one without repr?

src/parser/mod.rs
```diff
@@ -1,12 +1,73 @@
...
+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct ProgramVersionMismatch(u32, u32);
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct RPCVersionMismatch(u32, u32);
+
+#[allow(dead_code)]
+pub enum Arguments {
```
VadimYarovoy, line 50: Why one for NFSv3 and MOUNT, can it be separate structures? What we will have with NFSv4 in future? (not nearest future)

VadimYarovoy (changes requested):
First look, will look a second time in the evening to check tests and check logic algorithms

src/parser/parser_struct.rs
VadimYarovoy, line 1: result_i -> just result

src/parser/nfsv3/file.rs
```diff
@@ -0,0 +1,190 @@
...
+    #[test]
+    fn test_type_symlink() {
+        const DATA: &[u8] = &[0x00, 0x00, 0x00, 0x05];
+
+        let result = super::r#type(&mut Cursor::new(DATA)).unwrap();
+        assert!(matches!(result, file::Type::Symlink));
+    }
+
+    #[test]
+    fn tets_type_falure() {
```
VadimYarovoy: typo

src/parser/nfsv3/get_attr.rs
```diff
@@ -0,0 +1,33 @@
...
+#[cfg(test)]
+mod tests {
+    use std::io::Cursor;
+
+    use super::args;
+
+    #[test]
+    fn test_get_attr() {
+        #[rustfmt::skip]
+        #[rustfmt::skip]
```
VadimYarovoy, line 22: duplicate

VadimYarovoy (approved):

IgorErin (approved):

src/parser/nfsv3/create.rs
```diff
@@ -0,0 +1,141 @@
+//! Implements parsing for [`create::Args`] structure.
+
+use crate::parser::nfsv3::{file, MAX_FILENAME};
+use crate::parser::primitive::{array, option, string_max_size, u32, u64};
+use crate::parser::{Error, Result};
+use crate::vfs::create;
+use crate::vfs::create::{Verifier, VERIFY_LEN};
+use crate::vfs::file::Time;
+use crate::vfs::set_attr::{NewAttr, SetTime};
+use std::io::Read;
```
IgorErin, line 10: imports

PR #37, opened 2025-12-26
Mount protocol version 3 interface

### LINKED ISSUE #32: MNT

VadimYarovoy:
Resolves #32


src/mount/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+pub mod mnt;
+
+pub const MNTPATHLEN: usize = 1024;
+pub const MNTNAMLEN: usize = 255;
+pub const FHSIZE3: usize = 64;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Clone)]
+pub struct FileHandle(pub Vec<u8>);
```
artemiipatov: MOUNT file handle should correspond to NFS one (as in vfs::file in #21)

src/mount/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl FileHandle {
```
artemiipatov: I'm not sure if we should implement it in this PR.

src/mount/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Maximum bytes in a name.
+pub const MOUNT_HOST_NAME_LEN: usize = 255;
+/// Maximum bytes in a NFS v3 file handle.
+pub const HANDLE_SIZE: usize = 8;
+
+/// Unique file identifier.
+///
+/// Corresponds to the file handle from RFC 1813.
+#[derive(Clone)]
+pub struct Handle(pub [u8; HANDLE_SIZE]);
```
MikhailSvirukov: Maybe we should use [one](https://github.com/RMamonts/nfs-mamont/blob/0bad777245f5ba5788cce3ec910d37b2c1389772/src/vfs.rs#L85) from vfs or use this one everywhere

MikhailSvirukov: That also goes for all types in reference to Path and DirName

artemiipatov: We may use them, but soon it will be reworked in #21 and placed in NFS-protocol description. I thought, probably it would be nice to have isolated descriptions for each protocol. In any case, it can be changed easily.

IgorErin: I agree with @artemiipatov, separate for now

IgorErin (dismissed):

VadimYarovoy:
@IgorErin @artemiipatov

src/mount/umnt.rs
```diff
@@ -0,0 +1,28 @@
+//! Defines Mount version 3 [`Umnt`] interface (Procedure 3).
+//!
+//! as defined in RFC 1813 section 5.2.3.
+//! <https://datatracker.ietf.org/doc/html/rfc1813#section-5.2.3>.
+use async_trait::async_trait;
+
+use crate::vfs::FileName;
+
+pub type Result = std::result::Result<(), ()>;
```
artemiipatov: Result type with unit variants is unnecessary.

src/mount/export.rs
```diff
@@ -0,0 +1,33 @@
...
+
+/// Success result.
+pub struct Success {
+    /// Vector of export entries, each containing an exported
+    /// directory and a vector of clients that are allowed
+    /// to mount the specified directory.
+    pub exports: Vec<ExportEntry>,
+}
+
+pub type Result = std::result::Result<Success, ()>;
```
artemiipatov: Result type with unit fail variant is unnecessary, because it is supposed that there can be no other errors except nfs errors.

MikhailSvirukov (dismissed):

IgorErin (dismissed):

MikhailSvirukov (dismissed):

VadimYarovoy:
@IgorErin @MikhailSvirukov what we want to see here before merge?

dsemak (approved):

MikhailSvirukov (approved):

PR #52, opened 2026-02-15
fix: Validate consumed bytes in discard_current_message

dsemak:
Enhance the `discard_current_message` function to check if the consumed bytes exceed the current frame size, returning an error if so.


VadimYarovoy:
@dsemak, no issue or project link

src/parser/parser_struct.rs
```diff
@@ -355,8 +355,17 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
     /// Returns `Ok(())` if the message was successfully discarded, or an error
     /// if an I/O error occurs while discarding.
     async fn discard_current_message(&mut self) -> Result<()> {
-        let remaining = self.current_frame_size - self.buffer.total_bytes();
-        self.buffer.discard_bytes(remaining).await.map_err(Error::IO)?;
+        let already_consumed = self.buffer.total_bytes();
+        if already_consumed > self.current_frame_size {
+            return Err(Error::IO(io::Error::new(
```
VadimYarovoy, line 361: TO my mind it will be better to use not IO error, something like error enum with `FrameSizeExceeded` or `ProtocolViolation`

Or even existing `parser/mod.rs` - `MessageTypeMismatch`

src/parser/parser_struct.rs
```diff
@@ -355,8 +355,17 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
+                ErrorKind::InvalidData,
+                "Consumed more bytes than current frame size",
+            )));
+        }
+        let remaining = self.current_frame_size - already_consumed;
+        if remaining > 0 {
```
VadimYarovoy: useless check, `discard_bytes(0)` handles that case

src/parser/tests/parser_struct.rs
```diff
@@ -146,3 +146,21 @@ async fn parse_write_after_error() {
...
+        0x00, 0x00, 0x00, 0x00, // request
+        0x00, 0x00, 0x00, 0x05, // invalid rpc version (must be 2)
+    ];
+    let socket = MockSocket::new(buf.as_slice());
+    let alloc = MockAllocator::new(0);
+    let mut parser = RpcParser::new(socket, alloc, 0x20);
+
+    let result = parser.parse_message().await;
+    let error = result.err().unwrap();
+    assert!(matches!(error, Error::IO(io_err) if io_err.kind() == std::io::ErrorKind::InvalidData));
```
VadimYarovoy, line 165: strange test behavior, I mean that `proc_nested_errors` hides `RpcVersionMismatch` error and sends us just `InvalidData`

VadimYarovoy (changes requested):
Some controversial issues

src/parser/parser_struct.rs
dsemak: it is not obvious from the function description

src/parser/parser_struct.rs
dsemak: I would leave it as is and create a separate task, because I have questions about other places with IO errors

src/parser/parser_struct.rs
VadimYarovoy: agree with you, but nothings says that it is dangerous to pass 0

src/parser/parser_struct.rs
VadimYarovoy: ok

src/parser/tests/parser_struct.rs
MikhailSvirukov: As from #3 we split possible errors (currently represented in `parser::Error`, though i think they belong in `rpc`) as non-critical (ones, after that we just discard current message and continue receiving messages) and critical. I think that while discarding message we might get only one possible error (by `parser::Error`) - IO, which actually means that there are problems with server (or client) implementation. I think that we should send client that information, rather than any other, because it mean problems with implementation, rather than protocol

VadimYarovoy (dismissed):

MikhailSvirukov (approved):

MikhailSvirukov (dismissed):

src/parser/tests/parser_struct.rs
VadimYarovoy: ok

IgorErin (approved):

VadimYarovoy (approved):

PR #54, opened 2026-02-15
feat: Validate frame size in RPC parser

dsemak:
Add checks in the `parse_rpc_header` method to ensure that the frame size is not smaller than the XID width and does not exceed the maximum supported length. Add tests.


src/parser/parser_struct.rs
```diff
@@ -111,6 +111,19 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
         self.last = header & 0x8000_0000 != 0;
         self.current_frame_size = (header & 0x7FFF_FFFF) as usize;
 
+        if self.current_frame_size < std::mem::size_of::<u32>() {
+            return Err(Error::IO(io::Error::new(
+                ErrorKind::InvalidData,
+                "Frame size must include XID",
+            )));
+        }
+        if self.current_frame_size > MAX_MESSAGE_LEN {
```
MikhailSvirukov: Sorry, my bad, somewhere between multiple branch merges i lost this change. I fixed it in #55. So, it is not max message length, just inner buffer size. I think, that max size check come from next check (with `!self.lst`)

VadimYarovoy (dismissed):

src/parser/parser_struct.rs
VadimYarovoy: @MikhailSvirukov create task on it

VadimYarovoy (approved):

IgorErin (approved):

PR #55, opened 2026-02-15
parser: fix default inner buffer size

MikhailSvirukov:
Parser does use idea of several types limitations (filename and filepath), so that there would be some upper bounds to message sizes for each procedure. But though defined and calculated, this constant upper border wasn't included in #26


src/parser/parser_struct.rs
VadimYarovoy:
```suggestion
/// let mut parser = RpcParser::with_capacity(socket, alloc, 4096);
```

VadimYarovoy: or correct `RpcParser::new` usage here

VadimYarovoy (commented):
Review

IgorErin (dismissed):

VadimYarovoy (dismissed):

VadimYarovoy (approved):

IgorErin (approved):

PR #61, opened 2026-02-18
serialization: primitive types

MikhailSvirukov:
Serialization functions of primitive types by XDR

Resolves: #44

Include changes of #58 and #57

PR #62, opened 2026-02-18
serialization: procedures

MikhailSvirukov:
Small refactoring of RFC types, serialization functions for nfs and mount types

Include changes of #61


src/vfs/get_attr.rs
VadimYarovoy, line 1: for future, if you see that something is missed in proc, add it in separate pr, pr should be small as possible

src/serializer/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+/// All serialized data is aligned to [`ALIGNMENT`] (4 bytes) boundaries.
+pub const ALIGNMENT: usize = 4;
+
+/// Writes XDR alignment padding for an already-written field of length `n` bytes.
+///
+/// XDR requires 4-byte alignment; this emits zero bytes until the next multiple of [`ALIGNMENT`].
+fn padding(dest: &mut impl Write, n: usize) -> io::Result<()> {
+    let padding = (ALIGNMENT - n % ALIGNMENT) % ALIGNMENT;
+    let slice = [0u8; ALIGNMENT];
```
VadimYarovoy, line 35: make in module const

VadimYarovoy:
@MikhailSvirukov don't fix until we merge previous

VadimYarovoy:
@MikhailSvirukov time for this one

src/serializer/nfs/read_dir.rs
```diff
@@ -0,0 +1,37 @@
...
+    u64(dest, entry.file_id)?;
+    file_name(dest, entry.file_name)?;
+    u64(dest, entry.cookie.raw())
+}
+
+/// Serializes a list of [`Entry`] (linked list) into XDR.
+fn dir_list(dest: &mut impl Write, list: Vec<Entry>) -> io::Result<()> {
+    for e in list {
+        bool(dest, true)?;
+        entry(dest, e)?;
```
VadimYarovoy, line 22: Correct order and value?
```
 struct dirlist3 {
     entry3       *entries;
     bool         eof;
};
```

should it be ?

```
for e in list {
        entry(dest, e)?;
        bool(dest, false)?;
}
    bool(dest, true)

```

VadimYarovoy: same for read_dir_plus

VadimYarovoy (commented):
explain pls

src/serializer/nfs/read_dir.rs
MikhailSvirukov: Good point. I forgot `eof` flag. But current serialization seem to be ok: 
```
struct READDIR3resok {
           post_op_attr dir_attributes;
           cookieverf3  cookieverf;
           dirlist3     reply;
      };
``` 
resok takes single `dirlist3`; dirlist consists of linked list of `entry3` and single `eof` for whole list. So, cycle in `dir_list` serializes list of `entry3` (not list of `dirlist3` in XDR format - actually we might use `option`. 
I'll fix that)

VadimYarovoy (approved):

PR #64, opened 2026-02-24
fix(VFS): Remove unsupported file types from MkNode enum

### LINKED ISSUE #68: Extra types in mknoddata3
```
union mknoddata3 switch (ftype3 type) {
      case NF3CHR:
      case NF3BLK:
           devicedata3  device;
      case NF3SOCK:
      case NF3FIFO:
           sattr3       pipe_attributes;
      default:
           void;
 };
```

idea in comment - https://github.com/RMamonts/nfs-mamont/pull/64#discussion_r2851022296

VadimYarovoy:
according to
- RFC 1813
- man mknode

this PR is correct

file, dir, symlinc should not be created in mknode

default implementation now returns unsuported for all devices


nfs-mamont/src/vfs/mk_node.rs
```diff
@@ -17,12 +17,6 @@ pub enum What {
     Socket(NewAttr),
     /// Create fifo special file with specified initial attributes.
     Fifo(NewAttr),
-    /// Create regular file with no additional data.
-    Regular,
```
MikhailSvirukov, line 19: If i understand correctly union description (it is a bit ambiguous), than 
```
union mknoddata3 switch (ftype3 type) {
      case NF3CHR:
      case NF3BLK:
           devicedata3  device;
      case NF3SOCK:
      case NF3FIFO:
           sattr3       pipe_attributes;
      default:
           void;
 };
```
means, that all values of ftype3, that don't have arm go to default. That means that we can receive Regular file or Directory and apparently it is a correct request. Though it is unclear what to do than. Maybe protocol suggests we just ignore such messages?

VadimYarovoy: lets remove it for now because if we will fail in case - our error will be obvious. On the other hand  we can let server receive Regular file or Directory and that may lead to UB

created task for future - https://github.com/RMamonts/nfs-mamont/issues/68

IgorErin (changes requested):
I think, this place doesn't block us for now. Let's postpone this decision making about such cases for Vfs implementation.

VadimYarovoy:
@IgorErin @MikhailSvirukov 

ftype3 values ​​not listed in case fall into default (here - void, i.e. without additional payload). 
This does not make the query semantically correct.

For MKNOD only: NF3CHR, NF3BLK, NF3SOCK, NF3FIFO are legal. In other case, the server should return NFS3ERR_BADTYPE.

dsemak (approved):

PR #76, opened 2026-02-27
feat(VFS): change interface

### LINKED ISSUE #87: vfs: incorrect promise arg
look how others were implemented and change type to local Result type

```
/// Defines callback to pass [`GetAttr::get_attr`] result into.
#[async_trait]
pub trait Promise {
    async fn keep(attr: vfs::Result<file::Attr>);
}
```

VadimYarovoy:
Remove promise - we don't need it, and it will be difficult to implement for other users


VadimYarovoy:
@IgorErin @MikhailSvirukov review pls

MikhailSvirukov (approved):

PR #78, opened 2026-02-28
Splitting RPC errors from 

### LINKED ISSUE #72: New error types
Current version of `parser::Error` unifies all errors that can occur both from point of execution and protocol (RPC, mostly). Probably we should split them or narrow it down to protocol errors.

Initially, problem came up in #52, #57

MikhailSvirukov:
There is mix of errors we currently use in parsing incoming request: basic `io::Error`, some errors we introduce to show types limitations in our implementation of parser and RPC protocol-based errors. Current pull request suggests a way to resolve this mixture.

I look up in code that we use and came up with this: all parsing functions basically use single function inside - `io::Read::read_exact`, which returns either `ErrorKind::UnexpectedEof` or `io::Error`, used inside corresponding `Read` implementation. Since our implementations - `parser::CountBuffer` and `parser::ReadBuffer` - does not return any errors (logically - thier functions still return `io::Result`, and we don't want to use wrappers around `io::Error` ([here](https://github.com/RMamonts/nfs-mamont/pull/52#discussion_r2809076746)), we might map errors in such cases into single - `UnexpectedEof`. It might not look good, especially because it became dependent on other objects ( `parser::CountBuffer` and `parser::ReadBuffer` - they might be hard to modify or replace in the future) , but we do need to differ some types of `io::Error` - [here](https://github.com/RMamonts/nfs-mamont/blob/80f4fd737e2f764ab8fdd2f678af434639f34a5d/src/parser/read_buffer.rs#L121). 

Also, it is not exactly clear, how much detailed we want errors to be - RPC protocol suggests that 2 different errors are enough. 

And finally, i don't really like the way it was with module relations - parser and rpc became too coupled. Probably, it is natural thing, since we do parsing incrementally (and protocol defines it so), but may be there other ways to do so.

Resolves: #72

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,3 @@
 *secrets*.csv
 .env.local
 .env.production
-
-/fuzz
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@
 *secrets*.csv
 .env.local
 .env.production
+
+/fuzz
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10,10 +10,7 @@ mod rpc;
 #[cfg(test)]
 mod tests;
 
-use std::future::Future;
-
 use crate::parser::mount::{MountArgs, UnmountArgs};
-use crate::rpc::Error;
 use crate::vfs::{
     access, commit, create, fs_info, fs_stat, get_attr, link, lookup, mk_dir, mk_node, path_conf,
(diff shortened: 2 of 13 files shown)


MikhailSvirukov:
Discussed with @IgorErin and came to conclusion, that suggested variant does reveal inner implementation and became dependent on other object's implementations

PR #99, opened 2026-03-10
feat: Server context

### LINKED ISSUE #97: prototype of NFS context
Must store the following things:
- dyn vfs (for dispatch in vfs task)
- Export table (for mount, also will be used in vfs task)
- Allocator settings

VadimYarovoy:
Resolves https://github.com/RMamonts/nfs-mamont/issues/97

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use tokio::net::unix::SocketAddr;
 use tokio::sync::RwLock;
 
-use crate::mount::HostName;
+use crate::mount::{ExportEntry, MountEntry};
 use crate::vfs::{self, file};
 
 pub type SharedVfs = Arc<dyn vfs::Vfs + Send + Sync + 'static>;
@@ -17,16 +17,12 @@ pub struct ServerSettings {
     allocator_buffer_count: NonZeroUsize,
 }
 
-pub struct ServerExport {
-    allowed_hosts: Vec<HostName>,
-}
-
 struct ExportRegistry {
-    by_directory: HashMap<file::Path, ServerExport>,
+    by_directory: HashMap<file::Path, Vec<ExportEntry>>,
 }
 
 struct MountRegistry {
-    by_client: HashMap<SocketAddr, HashSet<file::Path>>,
+    by_client: HashMap<SocketAddr, HashSet<MountEntry>>,
 }
 
 pub struct ServerContext {
--- a/src/context.rs
+++ b/src/context.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use tokio::net::unix::SocketAddr;
 use tokio::sync::RwLock;
 
+use crate::allocator::Allocator;
 use crate::mount::{ExportEntry, MountEntry};
 use crate::vfs::{self, file};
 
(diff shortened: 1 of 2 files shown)


VadimYarovoy:
@MikhailSvirukov @dsemak ideas & suggestions?

src/context.rs
```diff
@@ -0,0 +1,32 @@
...
+    by_client: HashMap<SocketAddr, HashSet<MountEntry>>,
+}
+
+pub struct ServerContext<T: Allocator> {
+    allocator: T,
+    backend: SharedVfs,
+    // what's available to mount
+    exports: Arc<RwLock<ExportRegistry>>,
+    // who has mounted what
+    mounts: Arc<RwLock<MountRegistry>>,
```
MikhailSvirukov: So, this is a point for @ArtemNik1tin to start with Mount implementation?

MikhailSvirukov (dismissed):
Generally, looks alright) 
Probably, we should keep in mind (not in this pr, it's alright here, but generally), that in previous version all this shared data was lying to high on hierarchy, so it was really inconvenient to pass it, for example, access of mount table for mount handlers

src/context.rs
```diff
@@ -0,0 +1,32 @@
...
+    by_directory: HashMap<file::Path, ExportEntry>,
+}
+
+struct MountRegistry {
+    // one client can mount multiple dirs
+    by_client: HashMap<SocketAddr, HashSet<MountEntry>>,
+}
+
+pub struct ServerContext<T: Allocator> {
+    allocator: T,
```
VadimYarovoy: @MikhailSvirukov here is a question - parser takes allocator https://github.com/RMamonts/nfs-mamont/blob/main/src/parser/parser_struct.rs#L74-L100, we can replace `allocator` field here with global parser - or remain `allocator` and share it between parsers (parser per connection). Opinion?

src/context.rs
VadimYarovoy: Sure, @ArtemNik1tin you can find some inspiration in `next` branch and in our old implementation - https://gitlab.com/Yarovoy.vd/nfs-old and start implementation after we merge `ServerContext`

src/context.rs
MikhailSvirukov: I think better share `allocator` between connections, rather than parser. 
If i do understand currectly initial idea, there should be something like BufferPull (but current `Impl` and `Slice` do seem sufficient). Actually it looks like we can simply create `Impl` instance - it already has limitation sizes and `mpsc` needed - per connection. 
Global parser would probably be bottleneck, since we plan to handle several connections simultaneously.  I thought of it as one parser per connection (per `ReadTask`). 
So, maybe remove field `allocator` completely?

VadimYarovoy: if we are going to use allocator across conections, them maybe we should store `Arc<Mutex<Allocator>>` here? or from where global allocator will be accessed?

> Actually it looks like we can simply create Impl instance - it already has limitation sizes and mpsc needed - per connection.

limitations per connection sounds strange, if N connections have M buffers in global allocator, then M/N could be small value

VadimYarovoy: @MikhailSvirukov and in this schema - what whole server limits? what if there will be 1_000 clients?

VadimYarovoy: some idea, however global allocator sounds better - https://github.com/RMamonts/nfs-mamont/pull/105

VadimYarovoy:
@copilot+gpt-5.4 Analyze the structure of the parser and serializer, explain how they should be used, and in this regard, suggest how to create a server context with a shared resource limit for the entire server.

src/context.rs
VadimYarovoy: and global allocator with mutex here - https://github.com/RMamonts/nfs-mamont/pull/107

MikhailSvirukov (approved):

Issue #83, opened 2026-03-03
Cursor over Slice


MikhailSvirukov:
Currently, `allocator::Slice` does not implement any convenient way to read/write into it without explicitly iterating over it's buffers and handling read/write operations manually every time, though such operations might be very useful and frequently used

PR #100, opened 2026-03-10
feat(example): add mirrorfs example with full VFS implementation

### LINKED ISSUE #98: prototype of vfs impl
We need real example of fs, that uses tokio::fs

todo:
- implment all vfs ops
- write tests

VadimYarovoy:
This pull request introduces a new example filesystem, `mirrorfs`, which mirrors a local directory and implements the NFSv3 virtual filesystem (VFS) traits.

Implementation is probably bad, but it works

Resolves #98


VadimYarovoy:
@dsemak @ArtemNik1tin ideas & suggestions?

src/parser/parser_struct.rs
```diff
@@ -218,12 +222,34 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
     async fn parse_authentication(&mut self) -> Result<(OpaqueAuth, OpaqueAuth)> {
         let cred = self.buffer.parse_with_retry(auth).await?;
         let verf = self.buffer.parse_with_retry(auth).await?;
-        if !matches!(cred.flavor, AuthFlavor::None) || !cred.body.is_empty() {
+        let cred_ok = match cred.flavor {
+            AuthFlavor::None => cred.body.is_empty(),
+            AuthFlavor::Sys => true,
```
VadimYarovoy, line 228: maybe useless

src/service/mount/mod.rs
```diff
@@ -33,44 +32,68 @@ mod mnt;
...
 mod umntall;
 
-#[derive(Clone)]
-struct ExportPolicyEntry {
-    export: ExportEntry,
-    file_handle: file::Handle,
-    auth_flavors: Vec<AuthFlavor>,
-}
+// TODO: should be taken from config
+const AUTH: [AuthFlavor; 1] = [AuthFlavor::None];
```
VadimYarovoy, line 36: list of supported auth - should be here and not be passed by user

src/serializer/server/nfs/read_dir.rs
```diff
@@ -28,7 +28,8 @@ fn dir_list(dest: &mut impl Write, list: Vec<Entry>) -> io::Result<()> {
 pub fn result_ok(dest: &mut impl Write, arg: read_dir::Success) -> io::Result<()> {
     option(dest, arg.dir_attr, |attr, dest| file_attr(dest, &attr))?;
     array(dest, arg.cookie_verifier.raw())?;
-    dir_list(dest, arg.entries)
+    dir_list(dest, arg.entries)?;
+    bool(dest, arg.eof)
```
VadimYarovoy, line 32: @MikhailSvirukov missed eof?

was not working without it

src/serializer/server/nfs/read_dir_plus.rs
```diff
@@ -30,7 +30,8 @@ fn dir_list_plus(dest: &mut impl Write, list: Vec<Entry>) -> io::Result<()> {
 pub fn result_ok(dest: &mut impl Write, arg: read_dir_plus::Success) -> io::Result<()> {
     option(dest, arg.dir_attr, |attr, dest| file_attr(dest, &attr))?;
     array(dest, arg.cookie_verifier.raw())?;
-    dir_list_plus(dest, arg.entries)
+    dir_list_plus(dest, arg.entries)?;
+    bool(dest, arg.eof)
```
VadimYarovoy, line 34: @MikhailSvirukov missed eof?

was not working without it

src/serializer/server/serialize_struct.rs
```diff
@@ -332,11 +332,10 @@ impl<T: AsyncWrite + Unpin> WriteBuffer<T> {
     /// Flushes the staged XDR bytes followed by a streamed payload [`Slice`] (used for READ data).
     async fn send_inner_with_slice(&mut self, slice: Slice, count: usize) -> io::Result<()> {
         let slice_size = slice.iter().map(|b| b.len()).sum::<usize>();
-
-        if slice_size != count {
+        if count > slice_size {
```
VadimYarovoy: Not sure

scripts/nfs_manual_full_check.sh
VadimYarovoy, line 1: @MikhailSvirukov script to test basic behaviour

examples/mirror_fs/fs.rs
```diff
@@ -0,0 +1,332 @@
...
+    gid: None,
+    size: None,
+    atime: set_attr::SetTime::DontChange,
+    mtime: set_attr::SetTime::DontChange,
+};
+
+/// A file system implementation that mirrors a local directory.
+#[derive(Debug)]
+pub struct MirrorFS {
+    fsmap: Mutex<FsMap>,
```
VadimYarovoy: Mutex is bad idea

examples/mirror_fs/main.rs
```diff
@@ -0,0 +1,51 @@
...
+    eprintln!("mirrorfs listening on: {bind}");
+
+    let listener = TcpListener::bind(&bind).await?;
+    let export = ExportEntryWrapper {
+        export: ExportEntry {
+            directory: file::Path::new(export_root.to_string_lossy().into_owned()).unwrap(),
+            names: Vec::new(),
+        },
+        root_handle: root_handle,
+    };
```
VadimYarovoy: probably should be moved to cinstructor - not to show clien our implementation

VadimYarovoy (commented):
@MikhailSvirukov I have fixed that PR to ensure that everything works, we can try fo fix taht PR or add one more implementation which we will plan well (think we should move this way).

From this PR can see bugs and dix it inside `src/`

May be you will find something intresting here

HOW TO USE:
- prepare
```
mkdir /tmp/nfs-mount
mkdir /tmp/nfs-export
```
- run server
```
cargo run --example mirrorfs -- /tmp/nfs-export 0.0.0.0:2049
```
- mount
```
sudo mount -v -t nfs -o vers=3,tcp,proto=tcp,port=2049,mountport=2049,nolock 127.0.0.1:/tmp/nfs-export /tmp/nfs-mount
```
- ensure taht it works
```
chmod +x scripts/nfs_manual_full_check.sh
./scripts/nfs_manual_full_check.sh /tmp/nfs-mount/
```

examples/mirror_fs/fs.rs
VadimYarovoy: now is RwLock. My atttempt with lock free and dashmaps failed

VadimYarovoy:
output
```
mount.nfs: timeout set for Thu Mar 19 19:25:09 2026
mount.nfs: trying text-based options 'vers=3,tcp,proto=tcp,port=2049,mountport=2049,nolock,addr=127.0.0.1'
```

VadimYarovoy:
> output
> 
> ```
> mount.nfs: timeout set for Thu Mar 19 19:25:09 2026
> mount.nfs: trying text-based options 'vers=3,tcp,proto=tcp,port=2049,mountport=2049,nolock,addr=127.0.0.1'
> ```

@MikhailSvirukov its ok

futhermore, do not export sub dirs of exports is ok for our use case, so current mount service logic is good

VadimYarovoy:
read ones takes RwLock with write(), others are non blocking, most valuable for us are:
- read
- write

image

src/parser/parser_struct.rs
VadimYarovoy: not useless - can't work without it

src/serializer/server/serialize_struct.rs
```diff
@@ -331,15 +331,6 @@ impl<T: AsyncWrite + Unpin> WriteBuffer<T> {
 
     /// Flushes the staged XDR bytes followed by a streamed payload [`Slice`] (used for READ data).
     async fn send_inner_with_slice(&mut self, slice: Slice, count: usize) -> io::Result<()> {
-        let slice_size = slice.iter().map(|b| b.len()).sum::<usize>();
-
-        if slice_size != count {
```
MikhailSvirukov, line 336: @VadimYarovoy Did it backfire? Apparently it should be removed, because [there](https://github.com/RMamonts/nfs-mamont/blob/bbeb35054fbcab06619fa9536d673604990f7927/src/task/connection/vfs.rs#L63) we allocate `count` bytes, but if read procedure returns less (which is normal) check above would not pass, though everything is fine. 

Overall, it does not look problematic: yes, there must be some delay in other threads allocations if we request more buffers than we actually use, but that does not do any harm. Maybe, though we could add in `Slice` some field like `size` (probably not urgent - can be done when we get to #83 )?

VadimYarovoy: this check was conncted with mount problem

client
```
mount.nfs: timeout set for Fri Mar 20 16:21:17 2026
mount.nfs: trying text-based options 'vers=3,tcp,proto=tcp,port=2049,mountport=2049,nolock,addr=127.0.0.1'
mount.nfs: mount(2): Device or resource busy
```

i think we do #83 after this one, and many other optimizations

VadimYarovoy: or your question was about something other?

VadimYarovoy:
probably deadlock
```
fio --name=seq_write --rw=randrw --bs=1M --numjobs=10 --iodepth=16 --size=
1G --group_reporting --directory=/tmp/nfs-mount --ioengine=libaio
seq_write: (g=0): rw=randrw, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=libaio, iodepth=16
```

VadimYarovoy:
@MikhailSvirukov I think we should discuss here what should be chenged in `src/` and skip `examples/` for now, opinion?

MikhailSvirukov:
> @MikhailSvirukov I think we should discuss here what should be chenged in `src/` and skip `examples/` for now, opinion?

Well, i am a bit concerned that almost every module we have is public. And it looks like mostly it is cause of warnings in doc generation. Probably we can try minimize that amount

MikhailSvirukov:
> probably deadlock
> 
> ```
> fio --name=seq_write --rw=randrw --bs=1M --numjobs=10 --iodepth=16 --size=
> 1G --group_reporting --directory=/tmp/nfs-mount --ioengine=libaio
> seq_write: (g=0): rw=randrw, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=libaio, iodepth=16
> ```

@VadimYarovoy i think i understand why we have deadlock. Fio above does both reads and writes, and since `--iodepth=16`, it does not wait for replies and sends batches of requests. So, probably, when we parse read request and then several writes, but doing it before we start to process first read in vfs task (tokio does not switch to vfs_task before we receive all batch). In write requests we do exhaust allocator. So, then vfs_task starts to process requests in order, in which requests we received (that means in our case read and then all writes). But, if read requests allocation of memory, it would starts to wait for memory. But memory would be available only after we process writes, which come only after that read. That's deadlock. 

So, the problem is that we allocate in two tasks, which are sequential in terms of pipeline that we have. 

First solution, that comes to mind - do separate allocators for reads and writes.

VadimYarovoy:
> fio --name=seq_write --rw=randrw --bs=1M --numjobs=10 --iodepth=16 --size=
> 1G --group_reporting --directory=/tmp/nfs-mount --ioengine=libaio

cool, added second allocator and split their configs

dsemak (approved):

PR #104, opened 2026-03-10
parser: round-tests for proc args

MikhailSvirukov:
Current test coverage of parser functions, but it is quite impossible  to write unit tests with decent coverage. So current pull request introduce fuzz target to test such functions

PR #109, opened 2026-03-11
parser: fuzz tests

MikhailSvirukov:
Continuous testing for `parser::parser_struct`. 

Parser structure, that we currently use, lack of tests. But it is also quite hard to write unit-tests of any kind to test basic scenarios, because, firstly, it would require manually creating large binary data and secondly, parser is dependent on it's inner state, which is impossible to set, since we consider its state private ("black box"). That's why there is a crucial need for some kind of continuous data stream, that would test, whether parser can correctly maintain  it's inner state.


MikhailSvirukov:
For now i would just leave all fuzzing code in `fuzz` branch. Fuzz code got a bit spread between this and #104, so probably looking in fuzz branch would be more useful

VadimYarovoy:
> For now i would just leave all fuzzing code in `fuzz` branch. Fuzz code got a bit spread between this and #104, so probably looking in fuzz branch would be more useful

ok

PR #110, opened 2026-03-11
fix: bugs from parser

MikhailSvirukov:
parser fuzzing from #109 found two buds:
* invalid constant in test of maximum frame size; constant for default capacity of underlying buffer is used in place of actual size
* `rpc::Error::InvalidParsing` error cause parser fail, because it does not cause read retry in case, when more data need to be read before processing (quite rare cases when padding bytes we split between two read calls)

### DIFF
--- a/src/parser/nfsv3/file.rs
+++ b/src/parser/nfsv3/file.rs
@@ -220,7 +220,7 @@ mod tests {
     fn test_file_path_padding_error() {
         const DATA: &[u8] = &[0x00, 0x00, 0x00, 0x02, b'f', b'i', 0x00];
 
-        assert!(matches!(super::file_path(&mut Cursor::new(DATA)), Err(Error::IncorrectPadding)));
+        assert!(super::file_path(&mut Cursor::new(DATA)).is_err());
     }
 
     #[test]
--- a/src/parser/nfsv3/file.rs
+++ b/src/parser/nfsv3/file.rs
@@ -219,8 +219,7 @@ mod tests {
     #[test]
     fn test_file_path_padding_error() {
         const DATA: &[u8] = &[0x00, 0x00, 0x00, 0x02, b'f', b'i', 0x00];
-
-        assert!(super::file_path(&mut Cursor::new(DATA)).is_err());
+        assert!(matches!(super::file_path(&mut Cursor::new(DATA)), Err(Error::IO(_))));
     }
 
     #[test]
--- a/src/parser/parser_struct.rs
+++ b/src/parser/parser_struct.rs
@@ -146,12 +146,8 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
                 "Frame size must include XID",
             )));
         }
-        if self.current_frame_size > DEFAULT_SIZE {
-            return Err(Error::IO(io::Error::new(
-                ErrorKind::InvalidData,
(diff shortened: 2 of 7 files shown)


src/rpc.rs
```diff
@@ -86,8 +86,6 @@ pub enum Error {
     EnumDiscMismatch,
     /// An incorrect string was encountered during UTF-8 conversion.
     IncorrectString(FromUtf8Error),
-    /// Incorrect padding was found.
-    IncorrectPadding,
```
VadimYarovoy, line 90: IncorrectPadding error is impossible?

src/parser/parser_struct.rs
```diff
@@ -146,12 +146,8 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
             )));
         }
-        if self.current_frame_size > DEFAULT_SIZE {
-            return Err(Error::IO(io::Error::new(
-                ErrorKind::InvalidData,
-                "Frame exceeds maximum supported length",
-            )));
-        }
+
+        //TODO("there might max frame check, but with valid MAX_FRAME_SIZE")
```
VadimYarovoy: dont left TODO without issue links pls

src/parser/nfsv3/file.rs
```diff
@@ -220,7 +220,7 @@ mod tests {
     fn test_file_path_padding_error() {
         const DATA: &[u8] = &[0x00, 0x00, 0x00, 0x02, b'f', b'i', 0x00];
 
-        assert!(matches!(super::file_path(&mut Cursor::new(DATA)), Err(Error::IncorrectPadding)));
+        assert!(super::file_path(&mut Cursor::new(DATA)).is_err());
```
VadimYarovoy: `.is_err()` not really good for tests, what error is returned?

In general for now it is OK, but better use assert_eq() with expected error type

VadimYarovoy:
@MikhailSvirukov, this can be merged but first explain the comments above

src/rpc.rs
MikhailSvirukov: Actually it was used in single function - [padding](https://github.com/RMamonts/nfs-mamont/blob/fcfe440ae33eccab14b61ed4166bbaa56e146741/src/parser/primitive.rs#L14) in parsing. And bug showed up [there](https://github.com/RMamonts/nfs-mamont/blob/fcfe440ae33eccab14b61ed4166bbaa56e146741/src/parser/read_buffer.rs#L121). 

Basically, we should follow the rule - **if function, which does parsing, reach end of data, before finishing parsing, it should return `rpc::Error::IO(err)`, where `err.kind() == ErrorKind::UnexpectedEof`**. This is important for object on higher level to read more data, restore state and retry parsing. All functions does follow this rule except for `parser::padding` (actually, that's not completely true, since we use other errors like `rpc::Error::IncorrectString`, but they are considered critical and no retry is attempted). So, `Error::IncorrectPadding` shoots if padding bytes are separated between 2 (or more) read calls. It is highly unlikely that that actually happen, but there are no guarantees, how much bytes would `read` call from socket return, so it should be handled.

It is really hard to reproduce this error with raw byte data and mostly it came out because i tried to emulate some batched return from socket in `tests::socket` by adding `SEPARATE` constant. But fuzz did found it quite quickly

VadimYarovoy (approved):

Issue #137, opened 2026-03-14
Server verifier


MikhailSvirukov:
From `accepted_reply ` type from [RFC](https://datatracker.ietf.org/doc/html/rfc5531):
```
struct accepted_reply {
            opaque_auth verf;
            union switch (accept_stat stat) {
            ...
             } reply_data;
         };
```
`verf` - is an authentication verifier that the server generates in order to validate itself to the client.

As a part of authentication mechanism, server credentials/verifier should be provided. Probably such verifier data should be accessible from `WriteTask`  and passed (not accessed directly) from `serializer` to include such in reply

VadimYarovoy:
@MikhailSvirukov TODO or DONE?

VadimYarovoy:
@MikhailSvirukov can it be closed?

MikhailSvirukov:
> [@MikhailSvirukov](https://github.com/MikhailSvirukov) can it be closed?

Basically, it is empty now. I think we should leave it open for now (we also do have link in code on that issue)

PR #138, opened 2026-03-14
tasks: procedure results types

### LINKED ISSUE #131: channel type: Proc results
Channel should pass results of NFS protocol procedures (probably one from [serializer](https://github.com/RMamonts/nfs-mamont/blob/01aca003d5530e0efe8bcacbc1c6f805d2556cad/src/serializer/serialize_struct.rs#L89)) and specific information of RPC request (xid, probably auth info) from VFSTask to WriteTask

MikhailSvirukov:
Resolves #131

### DIFF
--- a/src/serializer/serialize_struct.rs
+++ b/src/serializer/serialize_struct.rs
@@ -11,8 +11,8 @@ use std::io::{ErrorKind, Write};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use crate::allocator::Slice;
-use crate::mount::{dump, export};
-use crate::rpc::{AcceptStat, Error, OpaqueAuth, RejectedReply, ReplyBody, RpcBody};
+use crate::mount::MountRes;
+use crate::rpc::{AcceptStat, Error, RejectedReply, ReplyBody, RpcBody};
 use crate::serializer::mount::mnt;
 use crate::serializer::nfs::{
     access, commit, create, error, fs_info, fs_stat, get_attr, link, lookup, mk_dir, mk_node,
@@ -21,8 +21,9 @@ use crate::serializer::nfs::{
 };
 use crate::serializer::rpc::auth;
 use crate::serializer::{u32, usize_as_u32, ALIGNMENT};
-use crate::vfs::STATUS_OK;
-use crate::{mount, serializer, vfs};
+use crate::task::{ProcResult, ReplyFromVfs};
+use crate::vfs::{NfsRes, STATUS_OK};
+use crate::{serializer, vfs};
 
 /// Minimum buffer size, that could hold complete RPC message
 /// with NFSv3 or Mount protocol replies, except for NFSv3 `READ` procedure reply -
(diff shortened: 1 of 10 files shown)


src/serializer/serialize_struct.rs
```diff
@@ -243,14 +193,21 @@ impl<T: AsyncWrite + Unpin> Serializer<T> {
...
-    pub async fn form_reply(&mut self, reply: ReplyFromVfs) -> io::Result<()> {
+    /// Serializes [`ProcReply`] into a complete XDR RPC reply and writes it to the underlying writer.
+    ///
+    /// ## Arguments:
+    /// *   `reply` - procedure result of [`ProcReply`] type
+    /// *   `verifier` - an authentication verifier of [`OpaqueAuth`] type that the server generates in
+    ///     order to validate itself to the client
+    ///
+    /// TODO:(<https://github.com/RMamonts/nfs-mamont/issues/137>)
+    pub async fn form_reply(&mut self, reply: ProcReply, verifier: OpaqueAuth) -> io::Result<()> {
```
VadimYarovoy, line 204: Explain why `verifier` not in `ProcReply`?

VadimYarovoy (commented):
Everything fine, but I have one question

src/vfs/mod.rs
```diff
@@ -175,3 +175,29 @@ pub trait Vfs:
     + commit::Commit
 {
 }
+
+/// Wrapper for all supported NFSv3 procedure result types coming from [`Vfs`].
+pub enum NfsRes {
```
VadimYarovoy, line 180: Why moved form `nfs3.rs` to `vfs/`?

VadimYarovoy: Maybe OK, then rename to `VfsRes`

src/serializer/serialize_struct.rs
VadimYarovoy: I don't mean to put it there, just need to logically understand whether it should be inside or not.

MikhailSvirukov: I described problem in #137. Apparently, auth info for that verifier is a property of server and not file system. And probably there is no need to move this data through all tasks, since it is a static info that does not change from server start.

src/vfs/mod.rs
MikhailSvirukov: It looks like it is more than some serialization-specific data. So it should be moved. I put it in `vfs` since we store most of nfs-related type there and it basically consists of types from there

src/serializer/serialize_struct.rs
MikhailSvirukov: Probably it should come from supplier of credential info. It depends on how we are going to organize Kerberos authentication. If it is a key-value storage in-memory, which is loaded on server start, probably server info will be somewhere around there. And if so, we could probably pass server info as arg to WriteTask

VadimYarovoy (approved):

VadimYarovoy (approved):

PR #141, opened 2026-03-14
refactor(parser): separate NFS and MOUNT procedure argument parsing

### LINKED ISSUE #140: fix: dividing a ProcArguments into parts
Looks like we don't need to unify NFS and MOUNT args, they are passed by different channels into their tasks

VadimYarovoy:
Resolves: #140

### DIFF
--- a/src/parser/parser_struct.rs
+++ b/src/parser/parser_struct.rs
@@ -72,7 +72,7 @@ const DEFAULT_SIZE: usize = 2500;
 ///
 /// # async fn example<A: Allocator, S: AsyncRead + Unpin>(socket: S, alloc: A) {
 /// let mut parser = RpcParser::new(socket, alloc);
-/// let args = parser.parse_nfs_message().await?;
+/// let args = parser.next_message().await?;
 /// # }
 /// ```
 pub struct RpcParser<A: Allocator, S: AsyncRead + Unpin> {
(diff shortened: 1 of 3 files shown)


VadimYarovoy:
@MikhailSvirukov ready for review

src/parser/parser_struct.rs
```diff
@@ -349,24 +287,70 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
         };
 
-        // that's done after normal parsing without errors
+        // finalize_parsing() is only called after successful header and procedure parsing; it is not run on error paths
         self.finalize_parsing()?;
         Ok(proc)
     }
 
+    /// Parses a complete MOUNT RPC message from the stream.
+    pub async fn parse_mount_message(&mut self) -> Result<MountArguments> {
```
MikhailSvirukov, line 331: Okay, it seems we can't that simply get rid of that enum. What function are we going to call from ReadTask? Apparently we need something like `next_message()` after all. Though let's keep the rest - i mean splitting parsing functions

VadimYarovoy: done

MikhailSvirukov (dismissed):

VadimYarovoy:
@MikhailSvirukov have fixed one more thing, now is ready, update fuzz branch according to these changes

MikhailSvirukov (approved):

PR #166, opened 2026-03-16
refactor: move vfs and mount to interface module

### LINKED ISSUE #164: move interfaces to interface module
move `vfs/` and `mount/` into new `interface/` module

or even in `api/` and move allocator trait here too

next step - split project to workspace - https://doc.rust-lang.org/cargo/reference/workspaces.html, present your vision of crates before implementation

VadimYarovoy:
Resolves: #164


VadimYarovoy:
not now

PR #175, opened 2026-03-16
feat(mount): impl mount service

### LINKED ISSUE #148: implement mount trait for MountService

VadimYarovoy:
Resolves: https://github.com/RMamonts/nfs-mamont/issues/148


VadimYarovoy:
@MikhailSvirukov here are some things to discuss:
- maybe we dont need mount trait :1st_place_medal: becase we need to pass too much inner things into procs (but now via args)
- we need vfs `obj` in mountservice becase we should somehow convert `dir` into `filehandle` by `path_to_handle` (new vfs trait)

src/service/mount/mnt.rs
```diff
@@ -1,14 +1,33 @@
...
+        let Ok(file_handle) = self.vfs.path_to_handle(&args.dirpath).await else {
+            return Err(Fail::NoEnt);
+        };
+
+        let mount_entry =
+            MountEntry { hostname: client_addr.ip().to_string(), directory: args.dirpath.clone() };
+
+        self.mounts.by_client.entry(client_addr).or_default().insert(mount_entry);
+
+        // TODO: take auth_flavors from config
```
VadimYarovoy: probably should be passed here (not in Args)

MikhailSvirukov:
> @MikhailSvirukov here are some things to discuss:
> 
>     * maybe we dont need mount trait 🥇 becase we need to pass too much inner things into procs (but now via args)
> 
>     * we need vfs `obj` in mountservice becase we should somehow convert `dir` into `filehandle` by `path_to_handle` (new vfs trait)

There are some thoughts i have on the matter:

- there are two mechanisms of filename to filehandle mapping resolution - with `mnt` from Mount and `lookup` from NFS; Mount initially solves problem of obtaining initial (probably root, but implemented file system might have more complex structure on the matter of paths, which it would give to client); so i believe that it is expected to client to act like that:

    1. client gets initial filehandle via Mount 
    2. client resolves all other paths in that directory with `lookup` (there is a problem, that client can't do `lookup` on higher level - like `ls ..`, but protocol [suggest](https://datatracker.ietf.org/doc/html/rfc1813#autoid-51) caching on client side or full path resolution
 
    Apparently this approach correlates with common access policies on large filesystems: client have access to single directory and does not have access to any directory from level above. 
So, if we suggest that client would work like that, than probably we don't need to resolve any filename from filesystem to filehandle (latter is a property of server - not file system, so we should not live mapping to vfs) in mount service. We only need to have mapping of for paths that actually can be mounted. Maybe we can create some list (mapping) as a part of server context 

- as for the mapping name resolution - it can be handled by vfs; basically the idea is to consider mapping that is handled by mount and lookup separate (though they have same mapping for root (or several paths that can be mounted)filehandles that are assigned  on server start)

- in vfs we might do lazy mapping - assign handle on request to path
- we can go even further: 

    > If two file handles from the same
   server are equal, they must refer to the same file, but if they
   are not equal, no conclusions can be drawn. Servers should try
   to maintain a one-to-one correspondence between file handles and
   files, but this is not required.
   
    which probably mean that we can do mapping resolution client-local as long as we can guarantee that there won't be same filehandle in other clients mapping (that can easily be achieved if we can assign unique number to every connection - all connections are handled from `handle_forever`)

VadimYarovoy:
@MikhailSvirukov did it in this spirit of https://github.com/RMamonts/nfs-mamont/pull/175#issuecomment-4070417292

VadimYarovoy:
Probably merge afret ReadTask

MikhailSvirukov (dismissed):
Everything looks alright. Probably we should write somewhere about this decision about mount structure somewhere

VadimYarovoy:
> Everything looks alright. Probably we should write somewhere about this decision about mount structure somewhere

sure, but i want to think about it until evening, there are some things that confuse me in current implementation

VadimYarovoy:
> Everything looks alright. Probably we should write somewhere about this decision about mount structure somewhere

@MikhailSvirukov ready

MikhailSvirukov (approved):

PR #177, opened 2026-03-17
tests: parser wrappers

### LINKED ISSUE #174: tests: wrapper results in parser
Currently we have introduced several wrappers, that return parser data from parser, but there are no checks in tests for header data (`xid` and `opaque_auth`), which should be fixed

MikhailSvirukov:
Resolves: #174

### DIFF
--- a/src/parser/tests/parser_struct.rs
+++ b/src/parser/tests/parser_struct.rs
@@ -1,14 +1,20 @@
-use num_traits::ToPrimitive;
-use std::sync::Arc;
-use tokio::sync::Mutex;
-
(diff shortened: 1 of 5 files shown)


src/parser/tests/parser_struct.rs
VadimYarovoy:
```suggestion
    push_u32(&mut payload, header.xid);
```

VadimYarovoy: change to
```rs
    push_u32(&mut payload, header.xid);
```

VadimYarovoy (approved):

PR #180, opened 2026-03-17
feat(task, vfs): implement vfs task

### LINKED ISSUE #122: Implement VfsTask

VadimYarovoy:
Resolves: https://github.com/RMamonts/nfs-mamont/issues/122


src/task/connection/vfs.rs
```diff
@@ -27,6 +33,53 @@ impl VfsTask {
...
+                }
+                NfsArguments::GetAttr(args) => NfsRes::GetAttr(self.backend.get_attr(args).await),
+                NfsArguments::SetAttr(args) => NfsRes::SetAttr(self.backend.set_attr(args).await),
+                NfsArguments::LookUp(args) => NfsRes::LookUp(self.backend.lookup(args).await),
+                NfsArguments::Access(args) => NfsRes::Access(self.backend.access(args).await),
+                NfsArguments::ReadLink(args) => {
+                    NfsRes::ReadLink(self.backend.read_link(args).await)
+                }
+                NfsArguments::Read(args) => NfsRes::Read(self.backend.read(args).await),
+                NfsArguments::Write(args) => NfsRes::Write(self.backend.write(args).await),
```
VadimYarovoy: @MikhailSvirukov are we going to use allocator here right now?

MikhailSvirukov: That's a hard part - somehow we want to force vfs implementation (user-defined) to use our allocator. Probably we could allocate and prepare Slice to write here and pass some kind of Write abstraction and make vfs implementation to write from read operation into this buffer. But again, make user use our Slice does not seem too bad. But probably we should not pass allocator (Impl)

MikhailSvirukov: Our maybe make Slice dynamic but with allocation size (like Vec)?

VadimYarovoy: added allocatr for read method

VadimYarovoy: @MikhailSvirukov or we are what to do something with write too (seems like everything ok wiht it)

MikhailSvirukov: Probably it is ok to pass Slice. Let's go with it

src/task/connection/vfs.rs
```diff
@@ -27,6 +42,68 @@ impl VfsTask {
...
+        let mut command_receiver = self.command_receiver;
+
+        while let Some(command) = command_receiver.recv().await {
+            let NfsArgWrapper { header, proc } = command;
+
+            let response = match *proc {
+                NfsArguments::Null => NfsRes::Null,
+                NfsArguments::GetAttr(args) => NfsRes::GetAttr(self.backend.get_attr(args).await),
+                NfsArguments::SetAttr(args) => NfsRes::SetAttr(self.backend.set_attr(args).await),
+                NfsArguments::LookUp(args) => NfsRes::LookUp(self.backend.lookup(args).await),
```
MikhailSvirukov, line 54: @VadimYarovoy  It occur to me, that some part of lookup procedure (probably the most informative one) - name resolution to filehandle should not happen in user-defined instance and probably user-defined file system should not be obliged to implement mapping. So probably we should not let user define it. 

I'll think, whether there are any other procedures that depend on server implementation (like lookup). 

Though... we would need that name resolution almost in every procedure. Okay, maybe we should do it one of two ways: pass name resolution server to vfs procedure and make client use it (sounds not good) or don't pass filehandle to procedure in vfs and resolve filehandle to name before calling vfs method. Second approach would rely on fact, that implemented vfs use name (path)  to identify object, but it seems as expected thing.

It is not a matter of this pr, simply should think of it when we start implementing name resolution

VadimYarovoy: good point, but
- In NFSv3, handle is the central essence of the protocol, and most operations already come with a handle, not a name.

- Even if the lookup is "taken to the server", we still need to keep the consistent handle model, stale handle, rename/hardlink semantics

- is not entirely accurate - many procedures go directly through the handle.
  > name resolution is needed in almost every procedure" 

So it seems to me ok to let backend solve all this stuff, but we can discuss it on status or earlier

MikhailSvirukov (approved):

PR #182, opened 2026-03-17
feat: implement read task

### LINKED ISSUE #120: Implement ReadTask

MikhailSvirukov:
Resolves #120

Merge after #180

### DIFF
--- a/src/task/connection/read.rs
+++ b/src/task/connection/read.rs
@@ -1,4 +1,3 @@
-use std::fmt::Debug;
 use std::io;
 use std::sync::Arc;
 
--- a/src/task/connection/read.rs
+++ b/src/task/connection/read.rs
@@ -1,23 +1,34 @@
+use std::fmt::Debug;
 use std::io;
+use std::sync::Arc;
 
-use crate::parser::NfsArgWrapper;
-use crate::task::global::mount::MountCommand;
-use crate::task::ProcReply;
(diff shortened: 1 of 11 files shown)


src/task/connection/read.rs
```diff
@@ -46,6 +52,67 @@ impl ReadTask {
...
+                // specific case when we couldn't parser xid, which means that we can't send reply
+                Err(ErrorWrapper { xid: None, .. }) => {
+                    return Err(io::Error::from(io::ErrorKind::Other));
+                }
+            }
+        }
     }
 }
+
+fn send_broken_pipe(
```
MikhailSvirukov, line 107: @VadimYarovoy i am not really sure about this, but probably we can try sending some information that error occur to client

VadimYarovoy: For a real BrokenPipe/write pipeline drop, the client is often already unavailable, so it's too late to send him something. But lets stay with it for now

It is more useful to add local diagnostics (log)  than to try to send the text to the client. Not in this PR

VadimYarovoy (dismissed):

src/task/connection/read.rs
VadimYarovoy: @MikhailSvirukov resolve this thread and merge when will be ready

VadimYarovoy (approved):

PR #192, opened 2026-03-23
allocator: two-level allocator

### LINKED ISSUE #191: Level allocator
Currently, there is a single allocator for all connections. Probably we could allow every connection to hold it's own small "cache" that would allow relatively small allocations to happen without request to global allocator.

MikhailSvirukov:
Resolves: #191 

This pr present multilevel allocator based on `Slice`. Though it pr presents it with 2 levels, it supports multiple by design.


MikhailSvirukov:
I haven't written any tests yet, since we are in a bit of hurry, but i did manage to run server and it seem to be working

MikhailSvirukov:
@VadimYarovoy i'll probably won't get things fixed before evening, so if you need something from here - you can do it right here. There are a few latest commits used to replace slice - probably we can reset them, i did them to run server and check, that things work

MikhailSvirukov:
Looks like there are two problems with that approach:
* it limits reads/writes from single connection that can be processed in parallel, since we request on upper level only if we get all buffers on local level
* on every connection creation new `Impl` is created, which causes multiple `Vec` allocations 

So, i think that there are better ways to implement multi-level allocation

VadimYarovoy:
@MikhailSvirukov i have already fixed allocator issue, will tell on Monday

PR #199, opened 2026-03-24
Unstable demo 

VadimYarovoy:
remove useless vector allocation in write proc of mirrorfs

futhermore used tokio::fs instead of std::fs in may cases

**this pr passed integration tests**

### DIFF
--- a/examples/mirror_fs/fs/read_impl.rs
+++ b/examples/mirror_fs/fs/read_impl.rs
@@ -1,6 +1,5 @@
-use std::io::SeekFrom;
-use tokio::fs::File;
-use tokio::io::{AsyncReadExt, AsyncSeekExt};
+use std::fs::File;
+use std::os::unix::fs::FileExt;
 
 use nfs_mamont::vfs::read;
 use nfs_mamont::Slice;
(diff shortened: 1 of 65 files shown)


VadimYarovoy:
@MikhailSvirukov @dsemak review pls

VadimYarovoy:
> @MikhailSvirukov @dsemak review pls

but do not merge? this PR will be for demo (presentation for public), later will fix all by small PRs in main

examples/mirror_fs/fs/read_impl.rs
```diff
@@ -17,27 +17,24 @@ impl read::Read for MirrorFS {
...
-                return Err(read::Fail {
-                    error: Self::io_error_to_vfs(&error),
-                    file_attr: Some(attr),
-                });
-            }
-        };
-
         let file_len = meta.len();
         let start = args.offset.min(file_len);
         let end = args.offset.saturating_add(args.count as u64).min(file_len);
```
MikhailSvirukov, line 33: Same as in `write` - check for READ_WRITE_MAX

examples/mirror_fs/fs/write_impl.rs
```diff
@@ -20,37 +20,57 @@ impl write::Write for MirrorFS {
...
                 wcc_data: Self::wcc_data(&path, before),
             });
         }
-        if let Err(error) = file.write_all(&data).await {
-            return Err(write::Fail {
-                error: Self::io_error_to_vfs(&error),
-                wcc_data: Self::wcc_data(&path, before),
-            });
+
+        let mut remaining = args.size as usize;
```
MikhailSvirukov: There we could check for READ_WRITE_MAX and reject too large requests. Probably it would be better to do that before allocation, but it raises module relations questions, so probably it could wait (looks like it does not do any harm right now)

Issue #214, opened 2026-03-31
vfs: task implmentation


MikhailSvirukov:
Service for handle resolution should be included in launching processing of each proc

PR #215, opened 2026-03-31
vfs: handle mapping integration into vfs task

### LINKED ISSUE #212: vfs: change interface signatures
Every `Args` structure from `vfs`, that represent corresponding procedure, should not take or return  any file::Handle instance, since it is a property of server and not user-defined implementation

MikhailSvirukov:
Current implementation does simple resolution from handle to map. It does not distinguish types of fs objects.
Introduced changes: 
* changed interfaces of `vfs` methods to avoid passing/returning `file::Handle` in `vfs` methods (they work with `Path`)
* methods of `vfs` work with absolute paths, when `HandleMap` implicitly stores relative

Resolves: #212 #214


examples/mirror_fs/fs.rs
```diff
@@ -1,20 +1,15 @@
...
 use nfs_mamont::consts::nfsv3::{NFS3_COOKIEVERFSIZE, NFS3_CREATEVERFSIZE};
 use nfs_mamont::vfs;
 use nfs_mamont::vfs::file;
 use nfs_mamont::vfs::read_dir;
 use nfs_mamont::vfs::set_attr;
 use nfs_mamont::vfs::write;
 use nfs_mamont::Slice;
-
-use crate::fs_map::FsMap;
+use std::fs::Metadata;
```
MikhailSvirukov, line 8: Reorder imports

examples/mirror_fs/fs.rs
```diff
@@ -52,23 +47,23 @@ const DEFAULT_SET_ATTR: set_attr::NewAttr = set_attr::NewAttr {
 /// A file system implementation that mirrors a local directory.
 #[derive(Debug)]
 pub struct MirrorFS {
-    fsmap: RwLock<FsMap>,
+    _root: PathBuf,
```
MikhailSvirukov: Do we need it? Also we create same root handle there and in MountServer separately - not good

src/task/connection/vfs.rs
```diff
@@ -42,23 +48,402 @@ impl VfsTask {
...
+                    }));
+                }
+                if let Err(error) = ensure_name_allowed(&args.from.name) {
+                    return NfsRes::Rename(Err(vfs::rename::Fail {
+                        error,
+                        from_dir_wcc: Self::default_wcc(),
+                        to_dir_wcc: Self::default_wcc(),
+                    }));
+                }
+
```
MikhailSvirukov: Possible deadlock? If concurrent "mirror" renaming

src/task/connection/vfs.rs
```diff
@@ -42,23 +48,402 @@ impl VfsTask {
...
+                let from_lock = from_dir.read().await;
+                let from = Self::join_name(from_lock.as_path(), args.from.name.as_str());
+
+                let to_lock = to_dir.read().await;
+                let to = Self::join_name(to_lock.as_path(), args.to.name.as_str());
+
```
MikhailSvirukov: There we should take lock on `from` and `to`  - first is being deleted and second just modified

src/task/connection/vfs.rs
```diff
@@ -42,23 +48,402 @@ impl VfsTask {
...
+                        NfsRes::Link(Ok(vfs::link::Success {
+                            file_attr: ok.file_attr,
+                            dir_wcc: ok.dir_wcc,
+                        }))
+                    }
+                }
+            }
+
+            NfsArguments::SymLink(args) => {
+                //TODO("check that path really exist?")
```
MikhailSvirukov: Probably concern of user implementation

src/task/connection/vfs.rs
```diff
@@ -72,44 +457,181 @@ impl VfsTask {
...
+                            },
+                        )
+                    }
+                }
             }
+
+            NfsArguments::FsStat(args) => match self.handles.path_for_handle(&args.root).await {
+                Err(error) => NfsRes::FsStat(Err(vfs::fs_stat::Fail { error, root_attr: None })),
+                Ok(lock) => {
+                    let path = lock.read().await;
```
MikhailSvirukov: Do we really need to lock root to get static info? Probably not

MikhailSvirukov:
Also we could think of a better hash function, so that entries of single directory would be in single shard (readdir would interfere with all other ops otherwise)

VadimYarovoy:
@MikhailSvirukov looks great will try to review at evening

MikhailSvirukov:
@VadimYarovoy there is also my concern about types that `vfs` procedures return. Most of them are fine, but several return `file::Handle`, which user fs implementation should not know about. Currently, it does not effect correctness, since these handles wrapped in Option (which is by RFC), and then properly replaced after, but we should not do that. This pr is quite big, since i didn't want to leave code in main, that does not work, but we can change result types of nfs procs in separate pr. What do you think?

src/handles.rs
```diff
@@ -0,0 +1,159 @@
...
+        let cached_source_paths = self.cached_paths_with_prefix(from_relative.as_path());
+        if !cached_source_paths.iter().any(|path| path == &from_relative) {
+            return Err(vfs::Error::StaleFile);
+        }
+
+        let overwritten_paths = self.cached_paths_with_prefix(to_relative.as_path());
+        for old_path in overwritten_paths {
+            if !cached_source_paths.iter().any(|source| source == &old_path) {
+                let full_old_path = self.to_full_path(old_path.as_path());
+                self.remove_path(full_old_path.as_path()).await?;
```
MikhailSvirukov: Actually, write locks should be acquired before deletion there, not in vfs task rename?

src/handles.rs
```diff
@@ -0,0 +1,159 @@
...
+        file::Handle(ROOT.to_be_bytes())
+    }
+
+    pub async fn path_for_handle(
+        &self,
+        handle: &file::Handle,
+    ) -> Result<Arc<RwLock<PathBuf>>, vfs::Error> {
+        let entry = self.handle_to_path.get(handle).ok_or(vfs::Error::StaleFile)?;
+        let relative = entry.value().read().await.clone();
+        Ok(Arc::new(RwLock::new(self.to_full_path(relative.as_path()))))
```
VadimYarovoy: rwlock on previous line clonned value - this will lead to datarace

src/task/connection/vfs.rs
```diff
@@ -42,23 +48,402 @@ impl VfsTask {
...
+                        from_dir_wcc: Self::default_wcc(),
+                        to_dir_wcc: Self::default_wcc(),
+                    }));
+                }
+
+                let from_lock = from_dir.read().await;
+                let from = Self::join_name(from_lock.as_path(), args.from.name.as_str());
+
+                let to_lock = to_dir.read().await;
+                let to = Self::join_name(to_lock.as_path(), args.to.name.as_str());
```
VadimYarovoy: something bad can happen between this two locks, rename?

src/handles.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub async fn create_handle(&self, path: &Path) -> Result<Handle, vfs::Error> {
+        let relative = self.relative_path(path)?;
+        if let Some(prev) = self.path_to_handle.get(&relative) {
+            return Ok(prev.value().clone());
+        }
+
+        let id = self.next_id.fetch_add(1, Ordering::Relaxed);
+        let handle = file::Handle(id.to_be_bytes());
+        self.handle_to_path.insert(handle.clone(), Arc::new(RwLock::new(relative.clone())));
+        self.path_to_handle.insert(relative.clone(), handle.clone());
```
VadimYarovoy, line 122: not attomic, can be bad state when first one is added and second not

Two call that simultaneously lookup at the same path will both receive None, generate two different handles, and write them - and only second will be saved (first lost)

src/handles.rs
```diff
@@ -0,0 +1,159 @@
...
+        for old_path in cached_source_paths {
+            let (_, handle) =
+                self.path_to_handle.remove(old_path.as_path()).ok_or(vfs::Error::StaleFile)?;
+            let new_path = Self::replace_path_prefix(
+                &old_path,
+                from_relative.as_path(),
+                to_relative.as_path(),
+            )?;
+
+            self.path_to_handle.insert(new_path.clone(), handle.clone());
```
VadimYarovoy: beetween remove old and insert new can be lookup that will return stale handle

src/handles.rs
```diff
@@ -0,0 +1,159 @@
...
+        if relative.as_os_str().is_empty() {
+            self.root.clone()
+        } else {
+            self.root.join(relative)
+        }
+    }
+
+    fn cached_paths_with_prefix(&self, prefix: &Path) -> Vec<PathBuf> {
+        self.path_to_handle
+            .iter()
```
VadimYarovoy: how big can be this map?

time will be always be O(N)

maybe Btree or store child handles links in parrent value

VadimYarovoy (commented):
Have some thoughts about

not sure that we need to change currect imterface in `src/vfs`. Some kind of middle layer can be added like

```
vfs_trait[1] (current) -> vfs_trait[2](new) -> mirrorfs
                                       |
                                       |
                           this one recieve all from [1]
                           calls to handle resolution and passed
                           to [2] and same logic in oposite way
```

futhermre find some bugs in current implementation

to my mind it is not obvious to add handle resolution logic and put it in work in same pr, such layer can be added in one PR and in some future one mirrorfs will be migrated on new interface

MikhailSvirukov:
So, currently:
* operations in `HandleMap`  are not atomic
* atomicity achieved by acquiring locks necessary for objects in `VfsTask`
* any object modification (rename, remove etc) locks parent directory
* rename (aka move) and remove/rmdir does not modify children paths

src/handles.rs
MikhailSvirukov: Probably we don't need it at all: implementing move/delete, where all objects in all sub directories also modified include semantic, that might be different from `vfs` implementation (when path, we move to is already exist, when dir to delete is not empty etc). I think we should keep this resolution as possible as we can

MikhailSvirukov:
> Have some thoughts about
> 
> not sure that we need to change currect imterface in `src/vfs`. Some kind of middle layer can be added like
> 
> ```
> vfs_trait[1] (current) -> vfs_trait[2](new) -> mirrorfs
>                                        |
>                                        |
>                            this one recieve all from [1]
>                            calls to handle resolution and passed
>                            to [2] and same logic in oposite way
> ```
> 
> futhermre find some bugs in current implementation
> 
> to my mind it is not obvious to add handle resolution logic and put it in work in same pr, such layer can be added in one PR and in some future one mirrorfs will be migrated on new interface

I'll make it another task. Maybe even before this pr, because there are plenty of boilerplate code in this pr, that could not be decrease because of different types (which are not actually).

VadimYarovoy:
> atomicity achieved by acquiring locks necessary for objects in VfsTask

NLM?

MikhailSvirukov:
> > atomicity achieved by acquiring locks necessary for objects in VfsTask
> 
> NLM?

Actually, we probably would not need this layer if we use NLM. Question is whether we could configure server, so it could communicate with clients that do requests both with and without NLM - that would probably be a problem. So no, this resolution is for `nolock` mode only

src/task/global/vfs.rs
```diff
@@ -104,22 +166,404 @@ impl VfsTask {
...
+                            Ok(handle) => handle,
+                            Err(error) => return FailRes::remove(error),
+                        };
+
+                        let lock = match self.handles.path_for_handle(&handle).await {
+                            Err(error) => return FailRes::remove(error),
+                            Ok(lock) => lock,
+                        };
+
+                        let _ = lock.write().await;
```
dsemak: But why?

dsemak (changes requested):

src/task/global/vfs.rs
MikhailSvirukov: Since we move handles out of vfs implementation, there should be some guaranties, that from procedures would reflect handles. Without lock inconsistency of fs and handles might happen. Most simple case is in `rename`: if `vfs` procedure did return successful result, but `HandleMap` service could not find handles (they were moved, deleted etc concurrently), which means, that we couldn't tell, what is a current state of fs. On one hand we do finish `vfs` proc (and fs knows about it), but by the time we get to handle updating state was changed - and there is no way of telling, how that should reflect on state of HandleMap.
So, locks are not required to keep some order of operations, which clients would see. They are required to keep `HandleMap` consistent with changes on `vfs` and `vfs proc + handle updating `must be atomic

PR #216, opened 2026-04-01
feat: implement sharded allocator

### LINKED ISSUE #162: allocator: sharded allocator
Now we have alobal allocator under `tokio::sync::Mutex`, we need to find smarter solution.

Some idea:
- in current allocator we have one channel - there could me more
- create wrapper with `alloc` method that takes channel by thread id (our index)
- if certain channel is empty - alloc can take slice from next channel (like hashmap conflicts resolution)
- maybe such slices should now where they need to be returned

VadimYarovoy:
Resolves: #162

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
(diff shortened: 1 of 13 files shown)


VadimYarovoy:
> do not merge, will be tested to find gives it us perf enefits or not

@MikhailSvirukov ready for review, zero-filling was removed due to perf improvement

MikhailSvirukov (approved):

PR #218, opened 2026-04-02
chore: remove self-written NFS functional check script

VadimYarovoy:
lets use nfstest as gold standart


VadimYarovoy:
@MikhailSvirukov i will provide instruction about how to use nfstest later, but lets merge this pr before (it is actually useless)

dsemak (approved):

PR #223, opened 2026-04-04
vfs: handle resolution

### LINKED ISSUE #213: vfs: define handle resolution interface

MikhailSvirukov:
This pr introduces `HandleMap` object, which is used for path resolution.
Several implementation details: 
* methods are not atomic on inner state modification, which means that there are no guaranties that between all three inner `DashMap` modifications in single method call no other thread wouldn't modify them
* rename is recursive - all files/dirs moved from one path to another would get new handles; if any intermediate dir/file already exists when we try to create one **whole subtree with root in this object is removed**
* rmdir, remove procedures does recursive update of handles; all handles which include directory removed, would be deleted
* names "", "." and ".." are not allowed   

Resolves #213


src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+        if let Some((_, children)) = self.directory_to_children.remove(from) {
+            self.directory_to_children.insert(to.to_path_buf(), children);
+        }
+
+        Ok(())
+    }
+
+    /// Converts a relative path into an absolute path under the configured root.
+    fn to_full_path(&self, relative: &Path) -> PathBuf {
+        if relative.as_os_str().is_empty() {
```
dsemak: this check is redundant

src/handles.rs
```diff
@@ -0,0 +1,400 @@
+#![allow(dead_code)]
```
dsemak: It is convenient to add dead_code to the lib.rs file under `mod handlers`

src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+    ///
+    /// Returns `StaleFile` if the path is unknown.
+    pub fn handle_for_path(&self, path: &Path) -> Result<Handle, vfs::Error> {
+        let entry = self.path_to_handle.get(path).ok_or(vfs::Error::StaleFile)?;
+        Ok(entry.value().clone())
+    }
+
+    /// Creates a handle for the given path if it does not already exist.
+    pub fn create_handle(&self, path: &Path) -> Result<Handle, vfs::Error> {
+        if let Some(prev) = self.path_to_handle.get(path) {
```
dsemak: Can be race condition between get (104 line) and insert (112 line), no atomicity

src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+        if relative.as_os_str().is_empty() {
+            self.root.clone()
+        } else {
+            self.root.join(relative)
+        }
+    }
+
+    /// Adds a child handle to a directory entry.
+    fn add_child_to_directory(&self, directory: &Path, handle: Handle) {
+        match self.directory_to_children.get_mut(directory) {
```
dsemak: The same race condition

src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+
+        Ok(handle)
+    }
+
+    /// Removes a path and its associated handle.
+    ///
+    /// # Non-recursive
+    /// Only the specific path is removed. Descendants are not touched.
+    pub fn remove_path(&self, path: &Path) -> Result<(), vfs::Error> {
+        let (_, handle) = self.path_to_handle.remove(path).ok_or(vfs::Error::StaleFile)?;
```
dsemak: We don't protect ROOT path

src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+
+        self.directory_to_children.remove(path);
+        Ok(())
+    }
+
+    /// Renames a path from `from` to `to`, updating all internal tables.
+    ///
+    /// # Non-recursive
+    /// Only the specific path is updated. Descendants are not rewritten.
+    pub fn rename_path(
```
dsemak: The same, we don't protect ROOT path

src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    /// Asserts that all internal tables of HandleMap match the expected state.
+    ///
+    /// `expected_paths` is a list of (path, handle) pairs that must exist.
+    /// This function checks:
+    /// - path_to_handle contains exactly these entries
+    /// - handle_to_path contains exactly these entries
+    /// - directory_to_children contains correct children sets
+    fn assert_state(map: &HandleMap, exp: &[(Handle, PathBuf)]) {
```
dsemak: The function checks that each path is contained in the children of its parent, but does not check the converse

src/handles.rs
```diff
@@ -0,0 +1,400 @@
...
+pub fn ensure_name_allowed(name: &file::Name) -> Result<(), vfs::Error> {
+    match name.as_str() {
+        "." => Err(vfs::Error::InvalidArgument),
+        ".." => Err(vfs::Error::Exist),
+        _ => Ok(()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dsemak: No test for rename with existing destination

Cargo.toml
```diff
@@ -23,6 +23,7 @@ tracing = { version = "0.1", features = ["release_max_level_off"] }
 tracing-subscriber = { version = "0.3", features = ["fmt", "env-filter"] }
 async-channel = "2.5.0"
 crossbeam-queue = "0.3.12"
+dashmap = "7.0.0-rc2"
```
dsemak: It's better to use the stable version

dsemak (changes requested):

MikhailSvirukov:
@dsemak 

- signatures might look a bit excessive, but they map args data from vfs of procedures, they will be used with
- it looks like there is no simple way to implement some kind of trie on paths, because we need resolution, where key is `file::Handle`. So, i don't really see, where we can use any-kind of tree-collection without hash-tables
- there is quite painful (for 2 reasons: firstly, we could be renaming big part of hashtable (since there is definitely hashtable - otherwise we can't resolve incoming handles) and secondly, it is not clear, what should we do in case we fail in renaming in `HandleMap` - `rename` in `vfs`  would commit changes anyway) to perform `remove`/`rename` operations, since we have dashmap, so i leave resolution from `Path` to `file::Handle`, so we could perform deletion of stale handles lazily; but this approach can lead to situation, where stale handles won't be deleted at all (probably it is critical problem)

MikhailSvirukov:
@dsemak i don't really like the way i do enforce taking locks _externally_ to make operations on _inner_ state atomic. Maybe we could think of integrating `vfs` directly inside `HandleMap`? That way sequence `resolve mapping - call vfs method - update mapping ` would happen in single function, not in several (current variant). Apparently it would become something close to `vfs` in sence ganesha make it (with pointers to methods of corresponding fs used). So, this would mean, that we receive request from VfsTask, call HandleMap method, method would collect locks, resolve handle, call method of `vfs`, update map and release locks. That way API would look more straight-forward - all methods would be atomic as user of `HandleMap` expects
Also it seems to me, it would make inner structure easier (we could remove bidirectional resolution - leave only handle-to-path)

VadimYarovoy (changes requested):
@MikhailSvirukov 

Today we found out that we should give nfs file handles in interface, because every backend wants to implement its own logic

Futher more in future (mean NFS4 and NLM) we should give locks outside to distributed systems, in case that server can be accessed from multiple ways

In general we should provide as low logic inside us as it possible

Sad to say but this PR has become unactual and should be closed 

For now our goal is too achive greater speed than ganesha to be usefull, probable tasks:
- rewrite mirrorfs without tokio at all and use uring
- look for nfs3 nvidia patches about multiconnect (custom nfs client is required)
- ... (will create tasks for things later)

PR #227, opened 2026-04-20
fix: turn on nagle's algorithm

### LINKED ISSUE #226: fix: disable nagle's algorithm
This algo slows us, we should wait TCP frame will be fully filled

VadimYarovoy:
in branch `demo` was discovered that without this option NFS works faster

Resolves: #226


VadimYarovoy:
@MikhailSvirukov @dsemak

MikhailSvirukov (approved):

PR #230, opened 2026-04-28
feat: vfs changes

VadimYarovoy:
- make VFS static
- Replace async-trait crate with trait-variant crate

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,17 +30,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-trait"
-version = "0.1.89"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -277,7 +266,6 @@ name = "nfs-mamont"
 version = "0.0.0"
 dependencies = [
  "async-channel",
- "async-trait",
  "byteorder",
  "crossbeam-queue",
  "num-derive",
@@ -286,6 +274,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "trait-variant",
 ]
 
 [[package]]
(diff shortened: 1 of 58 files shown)


VadimYarovoy:
@MikhailSvirukov, lets merge, i thought about: can we do it without extra deps and came to next

with trait we get siplier interface and no cons exept extra dep in cargo toml
#[trait_variant::make(Send)]
pub trait Access {
    /// Determines the access rights that a user, as identified by the credentials
    /// in the request, has with respect to a file system object.
    ///
    /// The results of this procedure are necessarily advisory in
    /// nature.  That is, a return status of [`Ok`] and the
    /// appropriate bit set in the bit mask does not imply that
    /// such access will be allowed to the file system object in
    /// the future, as access rights can be revoked by the server
    /// at any time.
    async fn access(&self, args: Args) -> Result<Success, Fail>;
}
```

This is more contraooled way that do the same but person who needs to  implement interface will be forsed to write async bloks inside their implementation
```
pub trait Access {
    /// Determines the access rights that a user, as identified by the credentials
    /// in the request, has with respect to a file system object.
    ///
    /// The results of this procedure are necessarily advisory in
    /// nature.  That is, a return status of [`Ok`] and the
    /// appropriate bit set in the bit mask does not imply that
    /// such access will be allowed to the file system object in
    /// the future, as access rights can be revoked by the server
    /// at any time.
    fn access(&self, args: Args) -> impl Future<Output = Result<Success, Fail>> + Send;
}

So lets stay with this trait for now

VadimYarovoy:
> @MikhailSvirukov, lets merge, i thought about: can we do it without extra deps and came to next
> 
> with trait we get siplier interface and no cons exept extra dep in cargo toml #[trait_variant::make(Send)] pub trait Access { /// Determines the access rights that a user, as identified by the credentials /// in the request, has with respect to a file system object. /// /// The results of this procedure are necessarily advisory in /// nature. That is, a return status of [`Ok`] and the /// appropriate bit set in the bit mask does not imply that /// such access will be allowed to the file system object in /// the future, as access rights can be revoked by the server /// at any time. async fn access(&self, args: Args) -> Result<Success, Fail>; }
> 
> ```
> 
> This is more contraooled way that do the same but person who needs to  implement interface will be forsed to write async bloks inside their implementation
> ```
> 
> pub trait Access { /// Determines the access rights that a user, as identified by the credentials /// in the request, has with respect to a file system object. /// /// The results of this procedure are necessarily advisory in /// nature. That is, a return status of [`Ok`] and the /// appropriate bit set in the bit mask does not imply that /// such access will be allowed to the file system object in /// the future, as access rights can be revoked by the server /// at any time. fn access(&self, args: Args) -> impl Future<Output = Result<Success, Fail>> + Send; }
> 
> So lets stay with this trait for now

cc @dsemak (you can approve too)

dsemak (approved):

PR #233, opened 2026-05-08
feat: integrate nix crate for filesystem statistics and enhance sizes of opts

VadimYarovoy:
- Start using real values insted of hardcoded ones for vfs stats
- Increase RW max to 1MB (like in knfs & ganesha) 
- Increase DIR_PREFETCH to 32 KB (like in knfs & ganesha)


VadimYarovoy:
@copilot fix conflicts

PR #236, opened 2026-05-10
feat: make mount service generic

VadimYarovoy:


src/lib.rs
```diff
@@ -56,35 +33,17 @@ pub fn init_tracing() {
...
-    let exports = exports
-        .into_iter()
-        .map(|entry| crate::service::mount::ExportEntryWrapper {
-            export: entry.export,
-            root_handle: entry.root_handle,
-        })
-        .collect();
-
-    let (mount_task, mount_sender) = MountTask::new(exports);
+    mount_service: M,
```
VadimYarovoy, line 40: @dsemak here or put it inside context?

VadimYarovoy:
@ArtemNik1tin nlm should follow same logic - there is some default service that we provide, but users can define thier own implementation

src/lib.rs
VadimYarovoy: for now - it fill be outside

MikhailSvirukov (approved):

PR #238, opened 2026-05-11
feat: pass link of mount service instead of its value

VadimYarovoy:
Now we pass link to MOUNT service, removed `&mut self` from Mount trait methods

### DIFF
--- a/src/service/mount/mod.rs
+++ b/src/service/mount/mod.rs
@@ -22,6 +22,8 @@
 use std::collections::{HashMap, HashSet};
 use std::net::SocketAddr;
 
+use tokio::sync::RwLock;
+
 use crate::mount::{ExportEntry, MountEntry};
 use crate::rpc::AuthFlavor;
 use crate::vfs::file;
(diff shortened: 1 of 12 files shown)


src/service/mount/mod.rs
```diff
@@ -80,17 +82,21 @@ struct MountRegistry {
 /// In-memory state backing the MOUNT v3 service implementation
 pub struct MountService {
     /// Exported directories that are available for mounting
-    exports: ExportRegistry,
+    exports: RwLock<ExportRegistry>,
     /// Active mounts keyed by client.
-    mounts: MountRegistry,
+    mounts: RwLock<MountRegistry>,
```
VadimYarovoy: need to check is it ok to have two rwlocks - can be inconsistent state in case of modification in other place?

VadimYarovoy: it is not providing interface to change inner fields, task by itselfs is only thing that works with moutservice, so evyrything is ok

VadimYarovoy:
@MikhailSvirukov @dsemak lets merge

dsemak (dismissed):

dsemak (approved):

PR #245, opened 2026-05-14
feat: inline small functions

VadimYarovoy:
should be testsed before merge


VadimYarovoy:
@MikhailSvirukov @dsemak review pls

MikhailSvirukov (approved):

PR #249, opened 2026-05-16
mount: add hostname struct

### LINKED ISSUE #248: Mount Hostname
Need to make struct for `mount::hostname` for better inside on it's size bounds (currently size bounds appear only in parser, which makes it not obvious that hostname is in fact  bounded string)

MikhailSvirukov:
Resolves #248

### DIFF
--- a/nfs-mamont/src/mount/mod.rs
+++ b/nfs-mamont/src/mount/mod.rs
@@ -6,10 +6,26 @@ pub mod mnt;
 pub mod umnt;
 pub mod umntall;
 
+use std::io;
+
+use crate::consts::mount::MOUNT_HOST_NAME_LEN;
 use crate::vfs::file;
 
 /// Client host name.
-pub type HostName = String;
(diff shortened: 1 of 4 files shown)


nfs-mamont/src/mount/mod.rs
```diff
@@ -6,10 +6,26 @@ pub mod mnt;
...
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
+pub struct HostName(String);
+
+impl HostName {
+    pub fn new(name: String) -> io::Result<Self> {
+        if name.len() > MOUNT_HOST_NAME_LEN {
+            return Err(io::Error::new(io::ErrorKind::InvalidInput, "host name too long"));
+        }
+        Ok(HostName(name))
+    }
```
VadimYarovoy, line 24: add line between functions

VadimYarovoy (approved):

PR #261, opened 2026-05-22
mirrorfs: readdir cache

MikhailSvirukov:
Resolves #258

### DIFF
--- a/mirror_fs/src/cache/readdir.rs
+++ b/mirror_fs/src/cache/readdir.rs
@@ -22,6 +22,7 @@ impl ReadDirCache {
             cache: Cache::builder()
                 .max_capacity(MAX_CACHE_CAPACITY)
                 .eviction_policy(EvictionPolicy::lru())
+                .support_invalidation_closures()
                 .build(),
         }
     }
@@ -30,8 +31,10 @@ impl ReadDirCache {
         self.cache.insert(parent.clone(), vec).await
     }
 
-    pub async fn invalidate_entry(&self, parent: &file::Handle) {
-        self.cache.invalidate(parent).await
+    pub fn invalidate_entry(&self, dir: PathBuf) {
+        self.cache.invalidate_entries_if(move |_, entry| {
+            entry.entries.iter().any(|(_, path)| path.starts_with(dir.as_path()))
+        });
     }
 
     pub async fn look_for_cache(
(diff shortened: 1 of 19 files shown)


mirror_fs/src/cache/readdir.rs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    /// Adds or updates cache entry for a directory
+    pub async fn add_entry(&self, parent: &file::Handle, vec: Arc<DirectoryListingSnapshot>) {
+        self.cache.insert(parent.clone(), vec).await
+    }
+
+    /// Invalidates all cache entries for directories containing files under the given path
+    pub fn invalidate_entry(&self, dir: PathBuf) {
```
MikhailSvirukov, line 44: How are we going to handle invalidation of cache in terms of transaction? Currently we need to go through all listings in `FsMap` and compare all prefixes. Probably, we don't need same approach with cache, since it would just add second equivalent layer.

mirror_fs/src/cache/readdir.rs
```diff
@@ -0,0 +1,58 @@
...
+        // what should we do in case of fail?
+        let _ = self.cache.invalidate_entries_if(move |_, entry| {
```
MikhailSvirukov, line 46: Not sure what should be done in case of fail? Invalidate everything?

mirror_fs/src/fs/read_dir_impl.rs
```diff
@@ -22,30 +26,46 @@ impl read_dir::ReadDir for MirrorFS {
...
         }
 
-        let entries = match self.list_directory_entries(&dir_path) {
-            Ok(entries) => entries,
-            Err(error) => return Err(read_dir::Fail { error, dir_attr: Some(dir_attr) }),
-        };
-
-        let total_entries = entries.len();
+        let entries =
+            if let Some(cached) = self.cache.read_dir_cache.look_for_cache(&args.dir).await {
```
MikhailSvirukov, line 30: Currently, we can't understand, whether parent of directory was invalidated

MikhailSvirukov (commented):
* i am not sure that we would gain much by cache for `readdir` - in case of modification (mkdir, create etc) we would need to iterate through whole `Fsmap` and cache.
* probably, we could do separate `Attr` cache with file granularity (not prefix)

@VadimYarovoy

VadimYarovoy:
@MikhailSvirukov close as not planned, can remain current branch (not delete it)

PR #263, opened 2026-05-22
feat: Buffer return type for allocate()

VadimYarovoy:
Method alloc now returns Buffer trait object insted of our inner Slice

Resolves https://github.com/RMamonts/nfs-mamont/issues/239

### DIFF
--- a/nfs-mamont/src/allocator/mod.rs
+++ b/nfs-mamont/src/allocator/mod.rs
@@ -15,27 +15,52 @@ use tokio::sync::Semaphore;
 
 pub use slice::Slice;
 
-type Buffer = Box<[u8]>;
+type PoolBuffer = Box<[u8]>;
 
 /// Shared state of the allocator to allow return of buffers and permit restoration.
 #[derive(Debug)]
 pub struct AllocatorState {
-    pub pool: ArrayQueue<Buffer>,
+    pub pool: ArrayQueue<PoolBuffer>,
     pub semaphore: Semaphore,
 }
 
-/// Allocates [`Slice`]'s.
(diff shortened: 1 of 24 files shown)


nfs-mamont/src/allocator/mod.rs
```diff
@@ -15,27 +15,52 @@ use tokio::sync::Semaphore;
...
-    /// Returns [`Slice`] of specified size.
+    /// Type of buffer returned by this allocator.
+    type Buffer: Buffer;
+
+    /// Returns a buffer of at least `size` bytes.
     ///
     /// # Parameters
     ///
-    /// - `size` --- size of returned slice.
+    /// - `size` --- minimum size of the returned buffer in bytes.
```
VadimYarovoy, line 71: fix doc, it is not min size

VadimYarovoy: it is min size, everything is ok

nfs-mamont/src/task/global/nlm.rs
```diff
@@ -8,46 +8,41 @@ use std::sync::Arc;
...
 where
+    B: Buffer + 'static,
     N: Nlm + Send + Sync + 'static,
 {
     /// Shared NLM service implementation.
     nlm_service: Arc<N>,
 
-    /// Channel for commands from client connection tasks.
-    receiver: UnboundedReceiver<NlmCommand>,
+    // Channel for commands from client connection tasks
```
VadimYarovoy:
```suggestion
    /// Channel for commands from client connection tasks
```

dsemak (approved):

PR #266, opened 2026-05-29
Add NLM Trait

ArtemNik1tin:
Implement the server-side NLM v4 (Network Lock Manager) protocol handlers
with an in-memory lock registry supporting shared/exclusive semantics and
range-based conflict detection.
Changes:
- Add `Lock`, `Unlock`, `Test`, `Cancel` traits with async Send support
  in `nlm::procedures`
- Implement `NlmService` with `LockRegistry` (HashMap<[u8;8], Vec<ActiveLock>>)
  and `ranges_overlap` conflict detection
- Add per-procedure implementations: lock, unlock, test, cancel
- Add XDR serializers for all NLM response types
- Make `NlmTask` generic over the NLM service, wire into `handle_forever`
- Update `mirror_fs` to create and pass `NlmService`

Resolves https://github.com/RMamonts/nfs-mamont/issues/235


nfs-mamont/src/service/nlm/cancel.rs
```diff
@@ -0,0 +1,103 @@
...
+use super::NlmService;
+
+impl Cancel for NlmService {
+    async fn cancel(&self, args: Nlm4CancelArgs) -> Nlm4CancelRes {
+        let fh_bytes = args.lock.file_handle.0;
+        let caller_name = args.lock.caller_name;
+        let system_identifier = args.lock.system_identifier;
+
+        let mut registry = self.locks.write().await;
+        registry.remove_by_owner(&fh_bytes, &caller_name, system_identifier);
```
VadimYarovoy: here can be removed multiple locks


- Process A (pid=100) acquires a lock on bytes [0–50)
- Process A (pid=100) acquires a lock on bytes [100–150)
- UNLOCK arrives with offset=0, length=50 (the goal is to release only the first one)
- `remove_by_owner` removes both locks, because filtering is only by `(caller_name, system_identifier)

possible solution - add range param to this fn

same for unlock

nfs-mamont/src/service/nlm/cancel.rs
```diff
@@ -0,0 +1,103 @@
+use crate::nlm::procedures::cancel::{Cancel, Nlm4CancelArgs, Nlm4CancelRes};
+use crate::nlm::Nlm4Stats;
+
+use super::NlmService;
+
+impl Cancel for NlmService {
+    async fn cancel(&self, args: Nlm4CancelArgs) -> Nlm4CancelRes {
```
VadimYarovoy, line 6: CANCEL == UNLOCK?

cancel shoul only be called for request that returned  `Nlm4Stats::Blocked`

but in that implementation answer is allways - GRANTED

nfs-mamont/src/service/nlm/unlock.rs
```diff
@@ -0,0 +1,101 @@
...
+use super::NlmService;
+
+impl Unlock for NlmService {
+    async fn unlock(&self, args: Nlm4UnlockArgs) -> Nlm4UnlockRes {
+        let fh_bytes = args.lock.file_handle.0;
+        let caller_name = args.lock.caller_name;
+        let system_identifier = args.lock.system_identifier;
+
+        let mut registry = self.locks.write().await;
+        registry.remove_by_owner(&fh_bytes, &caller_name, system_identifier);
```
VadimYarovoy: https://github.com/RMamonts/nfs-mamont/pull/266/changes#r3330805633

mirror_fs/src/main.rs
```diff
@@ -65,6 +65,7 @@ async fn main() -> std::io::Result<()> {
         });
     }
 
-    let mount_service = Arc::new(service::mount::MountService::with_exports(exports));
-    handle_forever(listener, context, mount_service).await
+    let mount_service = Arc::new(service::mount::MountService::with_exports(exports.clone()));
```
VadimYarovoy: for what .clone() was added?

nfs-mamont/src/serializer/server/nlm/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Serializes an [`Nlm4TestRes`] as the XDR reply body for `NLMPROC4_TEST`.
+///
+/// The reply is an XDR union: when the status is `Denied` the current
+/// lock holder information is included; otherwise only the status is
+/// written.
+pub fn test_res(dest: &mut impl Write, res: Nlm4TestRes) -> io::Result<()> {
+    cookie(dest, res.cookie)?;
+    stat(dest, res.test_stat.stat)?;
+    if let Some(holder) = res.test_stat.holder {
```
VadimYarovoy: Currently, the code writes the lock holder's data to the response if `holder.is_some()`, not looking on the value of `stat`. 

According to the XDR specification, `Nlm4TestReply` is a union: the `holder` field is only meaningful if `stat == Denied`.

 If someone creates `Nlm4TestReply { stat: Nlm4Stats::Granted, holder: Some(...) }`, the serializer will generate a squence that the XDR decoder on the client will read incorrectly.

However seems like current code works fine, so doublecheck it pls

Need to think what to do, suggestions are welcome

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -0,0 +1,138 @@
...
+    use crate::nlm::OpaqueHandle;
+    use crate::vfs::file::Handle;
+
+    fn handle(byte: u8) -> [u8; NFS3_FHSIZE] {
+        [byte; NFS3_FHSIZE]
+    }
+
+    fn opaque(val: u8) -> OpaqueHandle {
+        OpaqueHandle::new([val; OPAQUE_HANDLE_SIZE])
+    }
```
VadimYarovoy: same in all files

ArtemNik1tin:
@VadimYarovoy Everything is ready, I'm sorry that I didn't warn you that I won't be able to appear on the meetup today(

VadimYarovoy:
> @VadimYarovoy Everything is ready, I'm sorry that I didn't warn you that I won't be able to appear on the meetup today(

Hi, i was ill last two days, so at monday wasn't on meetup up too :disappointed:  (my coleege was instead of mee) Will look your PR untill week ends

ArtemNik1tin:
> > @VadimYarovoy Everything is ready, I'm sorry that I didn't warn you that I won't be able to appear on the meetup today(
> 
> Hi, i was ill last two days, so at monday wasn't on meetup up too 😞 (my coleege was instead of mee) Will look your PR untill week ends

Get well. Do you think it's possible to finish the work for this semester after we merge it?

VadimYarovoy:
> > > @VadimYarovoy Everything is ready, I'm sorry that I didn't warn you that I won't be able to appear on the meetup today(
> > 
> > 
> > Hi, i was ill last two days, so at monday wasn't on meetup up too 😞 (my coleege was instead of mee) Will look your PR untill week ends
> 
> Get well. Do you think it's possible to finish the work for this semester after we merge it?

by "work" you mean adding all nlm operations or something else?

ArtemNik1tin:
> > > > @VadimYarovoy Everything is ready, I'm sorry that I didn't warn you that I won't be able to appear on the meetup today(
> > > 
> > > 
> > > Hi, i was ill last two days, so at monday wasn't on meetup up too 😞 (my coleege was instead of mee) Will look your PR untill week ends
> > 
> > 
> > Get well. Do you think it's possible to finish the work for this semester after we merge it?
> 
> by "work" you mean adding all nlm operations or something else?

No, I meant "work" as a term paper. The task of adding the remaining operations will remain open.

nfs-mamont/src/serializer/server/nlm/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+/// written.
+pub fn test_res(dest: &mut impl Write, res: Nlm4TestRes) -> io::Result<()> {
+    cookie(dest, res.cookie)?;
+    stat(dest, res.test_stat.stat)?;
+    if res.test_stat.stat == Nlm4Stats::Denied {
+        let holder = res.test_stat.holder.expect("stat is Denied but holder is None");
+        u32(dest, holder.exclusive as u32)?;
+        u32(dest, holder.system_identifier as u32)?;
+        dest.write_all(holder.opaque_handle.as_bytes())?;
+        padding(dest, OPAQUE_HANDLE_SIZE)?;
```
VadimYarovoy: instead these two lines use 
```rs
vector(dest, holder.opaque_handle.as_bytes())?;
```

this will show you some incorect serializing in current test logic - missed XDR ln prefix

also cahnge
1060 -> 1064 in asserts

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,578 @@
...
+
+/// A blocked (pending) lock request waiting to be granted.
+struct PendingLock {
+    caller_name: String,
+    system_identifier: i32,
+    exclusive: bool,
+    offset: u64,
+    length: u64,
+    opaque_handle: OpaqueHandle,
+    #[allow(dead_code)]
```
VadimYarovoy, line 114: why dead code?

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,578 @@
...
+    cookie: Cookie,
+}
+
+/// In-memory collection of all active locks grouped by file handle.
+#[derive(Default)]
```
VadimYarovoy: as i see it is not used in main code, only in tests, better ad in test mod fn that creates default regestry

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,578 @@
...
+    /// Creates a new [`NlmService`] with the given export entries.
+    ///
+    /// The export list is accepted for API compatibility with the MOUNT
+    /// service but is not currently used for access control.
+    pub fn with_exports(_entries: Vec<ExportEntryWrapper>) -> Self {
+        Self::default()
+    }
+}
+
+#[cfg(test)]
```
VadimYarovoy: please dont use `#[cfg(test)]`, instead do next:
- create test molude
- define function or import that is only used in tests inside that module

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,578 @@
...
+mod tests {
+    use super::*;
+    use crate::nlm::cookie::Cookie;
+    use crate::nlm::lock::Nlm4Lock;
+    use crate::nlm::procedures::lock::{Lock, Nlm4LockArgs};
+    use crate::nlm::procedures::unlock::{Nlm4UnlockArgs, Unlock};
+    use crate::nlm::Nlm4Stats;
+    use crate::vfs::file::Handle;
+
+    // --- ranges_overlap ---
```
VadimYarovoy: such comments are not a good practice, to split tests by categories can be used:
- names
- test submodules

in that module you have a lot of tests, so i suggest you to move it i `test/` subdir, like in `serizalizer/server/tests/`

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -0,0 +1,175 @@
+use crate::nlm::procedures::lock::{Lock, Nlm4LockArgs, Nlm4LockRes};
+use crate::nlm::Nlm4Stats;
+
+use super::{ActiveLock, NlmService, PendingLock};
+
+impl Lock for NlmService {
+    async fn lock(&self, args: Nlm4LockArgs) -> Nlm4LockRes {
```
VadimYarovoy, line 7: not bug, but strange behavior

f a client already holds a range lock and requests one again, `find_conflict` doesn't check who owns the existing lock. It simply sees a range overlap and returns `Denied`.

Example:
- Client A holds an exclusive lock on `[0..100]`
- Client A requests an exclusive lock on `[0..100]` again
- Expected: `Granted`
- Got: `Denied`

VadimYarovoy: and one more thing, not bug, but need to think about

When client B sends a LOCK with block=true, and client A holds the lock, the server returns `Blocked` and stores the request in pending mode. When A UNLOCKs, grant_pending actually activates B's lock in by_file, but doesn't communicate this to B over the network.

According to RFC 1813, the server must send an asynchronous NLMPROC4_GRANTED (or NLMPROC4_GRANTED_MSG) to let client B know that its lock is now active. Without this, B could wait forever.

VadimYarovoy (commented):
Find some things to fix and others to think about

nfs-mamont/src/service/nlm/mod.rs
ArtemNik1tin: Stored here for the future granted callback. Currently unused because we don't yet send async callbacks. I plan to add this as a separate feature.

VadimYarovoy: Add todo comment with task link pls

ArtemNik1tin:
@VadimYarovoy I've corrected your comments and refactored a bit more, I hope it doesn't confuse you too much

nfs-mamont/src/service/nlm/tests/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use super::{ActiveLock, LockRegistry, NlmService};
+use crate::consts::nfsv3::NFS3_FHSIZE;
+use crate::nlm::cookie::Cookie;
+use crate::nlm::OpaqueHandle;
+
+mod ranges;
+mod registry;
+
+pub(super) fn fill_fh(value: u8) -> [u8; NFS3_FHSIZE] {
```
VadimYarovoy: seems like super is not needed, in functions below too

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -0,0 +1,215 @@
...
```
VadimYarovoy, line 7: not used args.reclaim in any logic

nfs-mamont/src/service/nlm/unlock.rs
```diff
@@ -0,0 +1,94 @@
...
+        ) {
+            Ok(new_lock) => new_lock,
+            Err(_) => return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed },
+        };
+
+        let mut registry = self.locks.write().await;
+
+        let fh_bytes = args.lock.file_handle.0;
+        registry.remove_by_owner(&fh_bytes, &target);
+        registry.grant_pending(&fh_bytes);
```
VadimYarovoy: why result of this function ignored? But problem is bigger, NLM tasks is kind special - it needs to sends not only results of just requested operations but also a callbacks

so now it doesn't work right

question for you, as i have no time untill wendsday on this week - is it a limitation of currently added operations of NLM without others or it can be fixed in that PR?

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+
+impl NlmService {
+    /// Creates an empty [`NlmService`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+#[cfg(test)]
+mod tests;
```
VadimYarovoy: all mods on the top

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        NlmService { locks: tokio::sync::RwLock::new(LockRegistry::new()) }
+    }
+}
+
+impl NlmService {
+    /// Creates an empty [`NlmService`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
```
VadimYarovoy, line 497: ok i see that default is needed in many tests, so you can 

- move all tests in files with same names in tests/operations/
- add fn (not impl) that creates default NlmService (will be seen only in tests module) in tests/mod.rs
- use this fn in tests/operations/lock.rs (and others not only lock.rs)

VadimYarovoy (commented):
short review, will look in tuesday/wendsday of week more

nfs-mamont/src/service/nlm/mod.rs
ArtemNik1tin: If there is no default implementation at all, then clippy warns. You can look CI to see that error.

VadimYarovoy: have seen that is used in main too. Then it my fault - return back `#[derive(Default)]`

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// and included in the returned vector. Requests that still conflict are
+    /// kept in the pending queue.
+    ///
+    /// ### Parameters
+    /// * `file_handle` — file whose pending queue should be rechecked.
+    ///
+    /// ### Returns
+    /// A vector of [`PendingLock`]s that have been granted —
+    /// the caller should send `NLMPROC4_GRANTED` for each one.
+    fn grant_pending(&mut self, file_handle: &[u8; NFS3_FHSIZE]) -> Vec<PendingLock> {
```
VadimYarovoy: grant_pending grants mutually-conflicting pending locks — doesn't check inter-pending conflicts

i will think more about how to handle it, but is it definatly a huge problem

nfs-mamont/src/serializer/server/nlm/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Serializes an [`Nlm4TestRes`] as the XDR reply body for `NLMPROC4_TEST`.
+///
+/// The reply is an XDR union: when the status is `Denied` the current
+/// lock holder information is included; otherwise only the status is
+/// written.
+pub fn test_res(dest: &mut impl Write, res: Nlm4TestRes) -> io::Result<()> {
+    cookie(dest, res.cookie)?;
+    stat(dest, res.test_stat.stat)?;
+    if res.test_stat.stat == Nlm4Stats::Denied {
+        let holder = res.test_stat.holder.expect("stat is Denied but holder is None");
```
VadimYarovoy: expect is very bad - stat: Denied` with `holder: None` will break this

VadimYarovoy: now it is save, but this will be bad in future

nfs-mamont/src/service/nlm/tests/operations/lock.rs
```diff
@@ -0,0 +1,196 @@
...
+    assert_eq!(res.stat, Nlm4Stats::Denied);
+}
+
+#[tokio::test]
+async fn lock_allows_shared_overlapping() {
+    let svc = NlmService::new();
+    svc.lock(make_lock_args_without_block(FH_DEFAULT, false, 0, LOCK_WHOLE_LENGTH, "test", 42, 0))
+        .await;
+    let res =
+        svc.lock(make_lock_args_without_block(FH_DEFAULT, false, 10, 20, "test", 42, 1)).await;
```
VadimYarovoy:
```suggestion
    let res = svc.lock(make_lock_args_without_block(1, false, 10, 20, "other", 99, 1)).await;
```

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+        if pending_requests.is_empty() {
+            self.pending.remove(file_handle);
+        }
+        has_request_been_deleted
+    }
+
+    /// Removes `target` from the active-lock list for `file_handle`.
+    /// Matching uses `PartialEq` (caller_name, system_identifier, offset, length).
+    fn remove_by_owner(&mut self, file_handle: &[u8; NFS3_FHSIZE], target: &ActiveLock) {
```
VadimYarovoy: The current implementation of **remove_by_owner** searches for an exact match of offset and length. 

To my mind (Check RFC for this), if the client has made a lock 0..100, and then sends unlock to 0.50, the lock should be cut (remain 50..100). 

Now she just doesn't go anywhere

VadimYarovoy: by the way why do not use this type?

```rs
/// Unique file identifier.
///
/// Corresponds to the file handle from RFC 1813.
#[derive(Debug, Clone, PartialEq)]
pub struct Handle(pub [u8; NFS3_FHSIZE]);
```

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -0,0 +1,39 @@
...
```
VadimYarovoy, line 7: When superimposing locks from the same client to the same range (e.g., an upgrade request from Shared to Exclusive), the new lock is definitely added to Vec<ActiveLock>. 
This will result in their accumulation without updating the type of locking or gluing ranges.

nfs-mamont/src/service/nlm/cancel.rs
```diff
@@ -0,0 +1,32 @@
+use super::{NlmService, PendingLock};
+use crate::nlm::procedures::cancel::{Cancel, Nlm4CancelArgs, Nlm4CancelRes};
+use crate::nlm::Nlm4Stats;
+
+impl Cancel for NlmService {
+    async fn cancel(&self, args: Nlm4CancelArgs) -> Nlm4CancelRes {
```
VadimYarovoy, line 6: Now cancel is looking for a request only in pending and returns the  Denied if it is not there.

According to the RFC, if the request has already managed to get Granted status (got into active_locks), the operation of the CANCEL should be ignored with the Granted response, so as not to break the client's static machine.

nfs-mamont/src/service/nlm/mod.rs
ArtemNik1tin: The RFC doesn't say anything about this, but this behavior is supported in POSIX locks and nfs-ganesha (https://github.com/nfs-ganesha/nfs-ganesha/blob/next/src/SAL/state_lock.c#L1102)

ArtemNik1tin: If anything, by "this" behavior, I mean what you said (if the client has made a lock 0..100, and then sends unlock to 0.50, the lock should be cut (remain 50..100))

ArtemNik1tin:
@VadimYarovoy I've corrected your comments, could you watch pr again?

VadimYarovoy:
> @VadimYarovoy I've corrected your comments, could you watch pr again?

Will look until Monday

nfs-mamont/src/task/global/nlm.rs
```diff
@@ -49,7 +79,26 @@ impl NlmTask {
 
             let nlm_result = match *proc {
                 NlmArguments::Null => NlmRes::Null,
-                _ => todo!(),
+                NlmArguments::Lock(nlm4_lock_args) => {
+                    debug!(xid = header.xid, "nlm task: proc=NLM");
```
VadimYarovoy: same line of logs on all procs

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -0,0 +1,40 @@
...
+            args.lock.opaque_handle,
+            args.cookie,
+        ) {
+            Ok(new_lock) => new_lock,
+            Err(_) => return Nlm4LockRes { cookie: args.cookie, stat: Nlm4Stats::Failed },
+        };
+
+        let fh = args.lock.file_handle;
+        if args.reclaim || registry.find_conflict(&fh, &ActiveLock::from(&new_lock)).is_none() {
+            if registry.push_or_replace(fh, ActiveLock::from(&new_lock)).is_err() {
```
VadimYarovoy: `ActiveLock::from(&new_lock)` called twice, looks like one time `ActiveLock::from(&new_lock)` is enough

nfs-mamont/src/service/nlm/unlock.rs
```diff
@@ -0,0 +1,32 @@
...
+            Err(_) => return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed },
+        };
+
+        let mut registry = self.locks.write().await;
+
+        let fh = args.lock.file_handle;
+        if registry.remove_by_owner(&fh, &target).is_err() {
+            return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed };
+        }
+        if registry.grant_pending(&fh).is_err() {
```
VadimYarovoy, line 28: add todo for future proc `NLMPROC4_GRANTED`

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+                "caller_name must not be empty",
+            ));
+        }
+
+        if caller_name.len() > nlm::LM_MAXSTRLEN {
+            return Err(Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!("caller_name is too long (max {})", nlm::LM_MAXSTRLEN),
+            ));
+        }
```
VadimYarovoy: same here and in PendingLock::new - move to fn

nfs-mamont/src/service/nlm/unlock.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::nlm::Nlm4Stats;
+
+use super::{ActiveLock, NlmService};
+
+impl Unlock for NlmService {
+    async fn unlock(&self, args: Nlm4UnlockArgs) -> Nlm4UnlockRes {
+        let target = match ActiveLock::new(
+            args.lock.caller_name,
+            args.lock.system_identifier,
+            true,
```
VadimYarovoy: why this value is hardcoded?

don't find mentions about it in rfc, if they are - give link pls

ArtemNik1tin:
@VadimYarovoy I've corrected your comments. I'm attaching the script as you requested
[test_nlm.py](https://github.com/user-attachments/files/28998002/test_nlm.py)

VadimYarovoy (approved):

PR #271, opened 2026-06-17
feat: mlock layout to speed up it first usage and avoid swap

VadimYarovoy:
Just mlock already known arena

maybe this should be a feature, not to force users run this with sudo or adding user to special group

### DIFF
--- a/nfs-mamont/src/allocator/mod.rs
+++ b/nfs-mamont/src/allocator/mod.rs
@@ -9,6 +9,7 @@ mod tests;
 
 use std::alloc::{self, Layout};
 use std::future::Future;
+#[cfg(feature = "mlock")]
 use std::io;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


nfs-mamont/src/allocator/mod.rs
```diff
@@ -101,10 +102,18 @@ impl Impl {
...
 
         let base_ptr = unsafe { alloc::alloc_zeroed(layout) };
+
         if base_ptr.is_null() {
             alloc::handle_alloc_error(layout);
         }
 
+        let ptr = base_ptr as *mut libc::c_void;
+
+        if unsafe { libc::mlock(ptr, total_size) } != 0 {
```
VadimYarovoy: maybe as feature?

nfs-mamont/src/allocator/mod.rs
dsemak, line 42: I think it would be better and clearer to add a numlock call

dsemak (approved):

PR #275, opened 2026-06-18
fuzzing

MikhailSvirukov:
There are 3 targets, introduced by this pr:
* parser target - one, that tests `RPCParser` struct with inputs, which doesn't make parser to fail and panic
* serializer - one, that tests `WriteBuffer`, that processes results for serialization
* round - simple roundtrips to ensure, that format of serialization matches format of parsing

Fuzzing includes all protocols, currently implemented - NFSv3, Mount and NLMv4

Also there are several mocks implemented and integrated in several tests (main reason for separate mod is that i also use them in fuzz)


nfs-mamont/src/mocks/buffer.rs
```diff
@@ -0,0 +1,196 @@
...
+        Self: Sized,
+    {
+        Self::new(vec![], 0)
+    }
+}
+
+#[cfg(feature = "arbitrary")]
+impl PartialEq for MockBuffers {
+    fn eq(&self, other: &Self) -> bool {
+        if self.range == other.range {
```
VadimYarovoy:
```suggestion
        if self.range != other.range {
```

maybe like that? if means else - then explain

nfs-mamont/src/mocks/read_socket.rs
```diff
@@ -0,0 +1,77 @@
...
+        _cx: &mut Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        let inner = self.get_mut();
+        if inner.end - inner.start == 0 {
+            inner.add_data();
+        }
+        let remaining_data = inner.end - inner.start;
+        let to_read = min(SEPARATE, min(buf.remaining(), remaining_data));
+        assert!(to_read > 0);
```
VadimYarovoy: Poll::Pending ?

nfs-mamont/src/nlm/procedures/test.rs
```diff
@@ -33,13 +35,27 @@ pub struct Nlm4TestRes {
...
 }
 
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for Nlm4TestReply {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        let stat = Nlm4Stats::arbitrary(u)?;
+        let holder = if stat == Nlm4Stats::Denied {
+            Some(Nlm4Holder::arbitrary(u)?)
+        } else {
+            Option::<Nlm4Holder>::arbitrary(u)?
```
VadimYarovoy: maybe just None? or can be other case?

nfs-mamont/src/mocks/mod.rs
VadimYarovoy, line 1: mocks folder here looks strange

nfs-mamont/src/nlm/procedures/test.rs
MikhailSvirukov: I think there might be Some too. Protocol only says some things about [denied](https://pubs.opengroup.org/onlinepubs/9629799/NLM_TEST.htm#tagcjh_11_17)

nfs-mamont/src/mocks/read_socket.rs
MikhailSvirukov: Apparently, rudimentary check. Thought it never came up in fuzzing (since there is always data to read), it might be a good idea to add just in case

nfs-mamont/src/mocks/buffer.rs
MikhailSvirukov: No, just bug. Apparently we don't need equality check , at all

nfs-mamont/src/mocks/mod.rs
MikhailSvirukov: Split and move to mods, they are used in

VadimYarovoy: i mean that  is a bit strange that lib reexports mocks, lib should only reexport Allocator trait in ideal, hovewer we gives an Impl too 

lets move them to fuzzing crate

MikhailSvirukov: I did it, just forget to push)

fuzz/fuzz_targets/parser_fuzz.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use libfuzzer_sys::fuzz_target;
+use nfs_mamont::allocator::mock::alloc::MockAllocator;
+use nfs_mamont::allocator::mock::buffer::MAX_BLOCK_AMOUNT;
+use nfs_mamont::parser::parser_struct::RpcParser;
+use nfs_mamont::parser::{NfsArguments, ProcArguments};
+use nfs_mamont::rpc::{Error, RpcBody, RPC_VERSION};
+use read_socket::FuzzMockSocket;
+use tokio::runtime::Runtime;
+use tokio::sync::Mutex;
```
VadimYarovoy, line 10: imports order

VadimYarovoy: in all files

nfs-mamont/src/allocator/mock/alloc.rs
```diff
@@ -0,0 +1,36 @@
+use std::num::NonZeroUsize;
+
+use crate::allocator::mock::buffer::MockBuffers;
+use crate::allocator::Allocator;
+
+pub struct MockAllocator {
```
VadimYarovoy, line 6: pub only with some feature, do not export it for general users

VadimYarovoy: for all things that had been made pub

nfs-mamont/src/vfs/read.rs
```diff
@@ -47,6 +51,26 @@ pub struct Args {
...
 
+#[cfg(feature = "arbitrary")]
+impl<'a, B> arbitrary::Arbitrary<'a> for Success<B>
+where
+    B: arbitrary::Arbitrary<'a> + Buffer,
+{
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        let data = B::arbitrary(u)?;
+        let count = data.len();
+        assert!(count < u32::MAX as usize);
```
VadimYarovoy: 1-4M is the biggest that can be passed through nfs3 protocol, 4 GB is prank, if you use something higher that 4M - mount will fail :)

nfs-mamont/src/vfs/write.rs
```diff
@@ -67,6 +74,20 @@ pub struct Args<B: Buffer> {
...
+        Ok(Self {
+            file: u.arbitrary()?,
+            offset: u.arbitrary()?,
+            size: data.len().to_u32().ok_or(arbitrary::Error::IncorrectFormat)?,
+            stable: u.arbitrary()?,
+            data,
+        })
+    }
+}
+
```
VadimYarovoy, line 91: maybe such imports should be moved to fuzz crate?

nfs-mamont/src/lib.rs
```diff
@@ -1,16 +1,15 @@
...
-mod parser;
-mod rpc;
-mod serializer;
+pub mod nlm;
+pub mod parser;
+pub mod rpc;
+pub mod serializer;
 pub mod service;
-mod task;
+pub mod task;
```
VadimYarovoy: why this pub?

nfs-mamont/src/lib.rs
```diff
@@ -1,16 +1,15 @@
 //! NFS Mamont - A Network File System (NFS) server implementation in Rust.
 
-mod allocator;
+pub mod allocator;
```
VadimYarovoy: is it nessesary to make mod pub, or it will be enough to re-export some inner entities under special feature?
```
pub use ...
```

fuzz/fuzz_targets/parser_wrapper.rs
```diff
@@ -0,0 +1,305 @@
...
+                }
+
+                NfsArguments::Commit(commit) => {
+                    commit::commit_args(&mut tmp_buffer, commit).unwrap()
+                }
+
+                NfsArguments::Null => (),
+            },
+
+            ProcArguments::Mount(mnt) => match *mnt {
```
VadimYarovoy: split match in functions by protocol, too big match

nfs-mamont/src/vfs/read.rs
VadimYarovoy: however, why assert and not error?
```rs
if count >= u32::MAX as usize {
    return Err(arbitrary::Error::IncorrectFormat);
}
```

VadimYarovoy (commented):
will be more

fuzz/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = { version = "0.4.10", features = ["arbitrary-derive"] }
+nfs-mamont = { path = "../nfs-mamont", features = ["arbitrary"] }
+arbitrary = "1.4.2"
+tokio = "1.49.0"
```
VadimYarovoy: tokio.workspace = true

fuzz/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+version = "0.0.0"
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = { version = "0.4.10", features = ["arbitrary-derive"] }
+nfs-mamont = { path = "../nfs-mamont", features = ["arbitrary"] }
```
VadimYarovoy: nfs-mamont.workspace = true

path looks uneeded, but i can be wrogn

nfs-mamont/src/vfs/write.rs
```diff
@@ -1,6 +1,8 @@
 //! Defines NFSv3 [`Write`] interface.
-
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 use num_derive::{FromPrimitive, ToPrimitive};
+use num_traits::ToPrimitive;
```
VadimYarovoy, line 6: #[cfg(feature = "arbitrary")] ?

fuzz/fuzz_targets/serializer.rs
```diff
@@ -0,0 +1,36 @@
...
+use nfs_mamont::task::ProcReply;
+use std::sync::OnceLock;
+use tokio::runtime::Runtime;
+use tokio::sync::Mutex;
+
+use write_socket::MockWriter;
+
+type TestSerializer = Serializer<MockBuffers, MockWriter>;
+static RUNTIME: OnceLock<Runtime> = OnceLock::new();
+static PARSER: OnceLock<Mutex<TestSerializer>> = OnceLock::new();
```
VadimYarovoy: SERIALIZER

fuzz/fuzz_targets/read_socket.rs
```diff
@@ -0,0 +1,79 @@
...
+impl AsyncRead for FuzzMockSocket {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        _cx: &mut Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        let inner = self.get_mut();
+        if inner.end - inner.start == 0 {
+            inner.add_data();
+        }
```
VadimYarovoy, line 64: _cx can cause problems with waker

```rs
if inner.end - inner.start == 0 {
    inner.start = 0;
    inner.end = 0;
    inner.data.clear();
    loop {
        match Pin::new(&mut inner.recv).poll_recv(cx) {
            Poll::Ready(Some(new_data)) => {
                inner.data.extend_from_slice(&new_data);
                inner.end = inner.data.len();
            }
            _ => break,
        }
    }
}
```

nfs-mamont/Cargo.toml
```diff
@@ -25,4 +26,7 @@ tracing-subscriber.workspace = true
 async-channel.workspace = true
 crossbeam-queue.workspace = true
 trait-variant.workspace = true
+
+arbitrary = { version = "1", optional = true, features = ["derive"] }
```
VadimYarovoy, line 30: can it be in fuzz crate?

VadimYarovoy: dev-dep

MikhailSvirukov:
@VadimYarovoy i managed to update this branch and check against new version of parser - everything seems to be okay. 
As for overall changes:
* there is one way, that i see, where we could move all `Arbitrary` from lib - if we do write manually all implementations (instead of deriving) inside fuzz crate; but that would mean many boilerlate code and making many things public; and there still should be feature, because fuzz uses nightly version of rust
* i create zero-allocator (always succeeds and really allocates nothing) - i did use it to detect memory leaks problems in fuzzing impl, but generally i don't use it - maybe it could be of some use with mock implementation of vfs to test overall pipeline processing (i mean logic with tasks and messages)?
* i created several macroses to do `pub use` for fuzz only with feature set
* also i do include parser in workspace

I am not sure, that we should merge it in main, because with cfg of `arbitrary` codebase looks messy, but i don't see a way to do it any better way (maybe you have any ideas?)

PR #277, opened 2026-06-22
xdr: explicit message size calculation

MikhailSvirukov:
Introduce trait, that calculates size of type in XDR format, and separate crate to allow deriving.

This is first step to solve #276 and #269 issues, but also it might be a good way to check inside serializer (maybe in fizz), that actual size of message matches expected

### DIFF
--- a/nfs-mamont/src/allocator/mod.rs
+++ b/nfs-mamont/src/allocator/mod.rs
@@ -15,6 +15,8 @@ use std::sync::Arc;
 use crossbeam_queue::ArrayQueue;
 use tokio::sync::Semaphore;
 
+use crate::xdr;
+
 pub use buffer::UnownedBuffer;
 pub use slice::Slice;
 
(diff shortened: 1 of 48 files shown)


nfs-mamont/src/allocator/mod.rs
```diff
@@ -40,7 +42,7 @@ impl Drop for AllocatorState {
 /// Abstract buffer type returned by [`Allocator`].
 ///
 /// Implementations provide chunked read/write access to the allocated memory.
-pub trait Buffer: Send + Sync {
+pub trait Buffer: Send + Sync + xdr::XDRSize {
```
MikhailSvirukov, line 45: Probably should not be there

nfs-mamont/src/vfs/mod.rs
```diff
@@ -227,3 +228,96 @@ pub enum NfsRes<B: Buffer> {
     PathConf(std::result::Result<path_conf::Success, path_conf::Fail>),
     Commit(std::result::Result<commit::Success, commit::Fail>),
 }
+
+impl<B: Buffer> xdr::XDRSize for NfsRes<B> {
```
MikhailSvirukov, line 232: Or should i rewrite derive to eliminate discriminant counting in enum?

PR #278, opened 2026-06-25
feat: vectorized write in mirrorfs

VadimYarovoy:
Before this pr we had extra copy of data to write it with just one syscall - write_all, but we dont actually need extra copy for that, so i just replaced write_all with vectorized call

### DIFF
--- a/mirror_fs/src/fs/mod.rs
+++ b/mirror_fs/src/fs/mod.rs
@@ -12,7 +12,6 @@ use nfs_mamont::vfs::file;
 use nfs_mamont::vfs::read_dir;
 use nfs_mamont::vfs::set_attr;
 use nfs_mamont::vfs::write;
-use nfs_mamont::Buffer;
 
 use crate::fs_map::FsMap;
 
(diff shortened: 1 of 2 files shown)


VadimYarovoy:
@dsemak @MikhailSvirukov review pls

mirror_fs/src/fs/write_impl.rs
VadimYarovoy, line 73: is it correct to use it instead of flush?

mirror_fs/src/fs/write_impl.rs
```diff
@@ -37,19 +37,24 @@ impl<B: Buffer> write::Write<B> for MirrorFS {
...
             });
         }
-        if let Err(error) = file.write_all(&data).await {
-            return Err(write::Fail {
-                error: Self::io_error_to_vfs(&error),
-                wcc_data: Self::wcc_data(&path, before),
-            });
-        }
+
+        let size = match file.write_vectored(&data).await {
```
dsemak, line 60: write_vectored doesn't always write data to the end and may return n < total

mirror_fs/src/fs/mod.rs
```diff
@@ -208,18 +207,6 @@ impl MirrorFS {
         }
     }
 
-    fn collect_buffer_bytes(slice: &impl Buffer, size: u32) -> Vec<u8> {
```
dsemak, line 211: This code truncated the data to args.size, but the new code writes all buffer chunks.
Non critical, but we need to determine what behavior we want.

dsemak (changes requested):

mirror_fs/src/fs/write_impl.rs
VadimYarovoy: according to RFC

> count
         The number of bytes of data written to the file. The
         server may write fewer bytes than requested. If so, the
         actual number of bytes written starting at location,
         offset, is returned.

and here is the value, so behavior is OK - https://github.com/RMamonts/nfs-mamont/pull/278/changes#diff-eda6b99b0bd33fcefa247af882df680a8650492877f83575d222417ec3c01181R72

dsemak (approved):

MikhailSvirukov (approved):

PR #285, opened 2026-07-22
Next

VadimYarovoy:
- tcp nodelay
- single allocator
- parser simplification
- sys auth support

### DIFF
--- a/nfs-mamont/src/parser/read_buffer.rs
+++ b/nfs-mamont/src/parser/read_buffer.rs
@@ -1,362 +1,190 @@
-//! Buffered reading utilities for parsing XDR-encoded RPC messages.
+//! Buffered frame reading for parsing XDR-encoded RPC messages.
 //!
-//! This module provides a double-buffered reader that efficiently handles
-//! asynchronous reading from network sockets while supporting retry logic
-//! for parsing operations that may require additional data.
+//! This module provides [`FrameReader`] --- a buffered reader that exposes the
+//! bodies of RMS frames (RFC 5531, section 11) of an async stream to
+//! synchronous parsing code.
 //!
-//! The main component is [`CountBuffer`], which wraps an [`AsyncRead`] stream
-//! and provides a [`Read`] interface for synchronous parsing functions. It uses
-//! two internal buffers to allow reading new data while
-//! still being able to retry parsing from a previous position if needed.
+//! The reader guarantees that after [`FrameReader::begin_body`] the first
+//! `min(frame_size, capacity)` bytes of the frame body (the "head window") are
+//! buffered in memory, so procedure arguments are parsed synchronously via the
+//! [`Read`] implementation without any retries or awaits. Only opaque payloads
+//! that do not fit into the head window (NFSv3 `WRITE` data) are read from the
+//! socket directly via [`FrameReader::read_body_exact`].
+//!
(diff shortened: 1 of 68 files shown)


nfs-mamont/src/task/connection/write.rs
```diff
@@ -42,20 +48,38 @@ impl<B: Buffer> WriteTask<B> {
...
+                batched += 1;
+
+                if batched < MAX_BATCH_REPLIES && serializer.buffered_len() < FLUSH_THRESHOLD {
+                    next = result_receiver.try_recv().ok();
+                }
+            }
+
+            if let Err(e) = serializer.flush().await {
+                error!(error=%e, "write task: failed to send replies");
+                serializer.reset();
```
MikhailSvirukov, line 81: Probably, we do not need explicit reset here. `WriteBuffer::flush`, which is called in sequence of calles after `Serializer::flush` already does that

nfs-mamont/src/task/connection/write.rs
```diff
@@ -42,20 +48,38 @@ impl<B: Buffer> WriteTask<B> {
...
+                    // corrupted, but keep the complete replies staged before it.
                     // TODO: Consider closing connection or continuing based on error type
                     // For now, continue processing other replies
+                    serializer.truncate_staged(staged);
                 }
-            };
+                batched += 1;
+
+                if batched < MAX_BATCH_REPLIES && serializer.buffered_len() < FLUSH_THRESHOLD {
+                    next = result_receiver.try_recv().ok();
```
MikhailSvirukov, line 75: It looks alright, though i have some concerns, if it would not cause troubles. Batching is a good solution in case we always have plenty of responses (which is our basic scenario). In case of rare messages (when producer does not manage to send responses quickly enough), this works as "timeout", so response does not stay in buffered stage for too long. But what if we simply has large requests? And ultimately consumer (write task) can ask for replies faster, than requests are processed. Question is, if such replies still need to be staged?

VadimYarovoy: need to to be tested

VadimYarovoy:
@MikhailSvirukov this will be not merged (should be draft) its just branch for experiments

nfs-mamont/src/parser/read_buffer.rs
MikhailSvirukov, line 1: As i understand it - it looks all right. You simply change two buffers and retry logic to explicit buffer filling. Probably it is even better solution, since there is no repeated parsing on not-fully-received data (but probably they do not happen a lot, since header sizes are really small in most cases, so maybe i overthink that).

MikhailSvirukov (commented):
Parsing is okay, maybe we could do even auth part - it simply change interfaces

nfs-mamont/src/parser/read_buffer.rs
MikhailSvirukov: And also, i think it would be a good thing to run fuzzer on it. I'll try to do it today

VadimYarovoy: can you move this part as PR? I am now thinking about allocator stuff

VadimYarovoy: and create tasks for it (issues)

VadimYarovoy:
dead, possibly we move all we needed from here

PR #286, opened 2026-07-23
Implement NLMPROC4_GRANTED callback for blocking lock requests

ArtemNik1tin:
Per RFC 1813, when a server responds with `NLM4_BLOCKED` to a blocking lock request, it must send an asynchronous `NLMPROC4_GRANTED` callback RPC once the lock becomes available. Currently, `grant_pending()` returns the newly-granted locks but the caller discards the result.

This PR adds:

- RPC callback channel (`granted_tx`/`granted_rx`) per connection for server-initiated `NLMPROC4_GRANTED`
- Cookie delivery via channel in `unlock.rs` after `grant_pending()`
- RPC CALL serialization in `Serializer::form_call()`
- `tokio::select!` in `WriteTask` for multiplexing replies and callbacks
- Tests for serialization, callback flow, and no-callback path

Resolves https://github.com/RMamonts/nfs-mamont/issues/267

### DIFF
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -237,7 +237,7 @@ impl<B: Buffer, T: AsyncWrite + Unpin> Serializer<B, T> {
         cred: OpaqueAuth,
         verifier: OpaqueAuth,
     ) -> io::Result<()> {
-        u32(&mut self.buffer, rand::random::<u32>())?;
+        u32(&mut self.buffer, SystemTime::now().duration_since(UNIX_EPOCH))?;
         u32(&mut self.buffer, RpcBody::Call as u32)?;
         u32(&mut self.buffer, RPC_VERSION)?;
         u32(&mut self.buffer, NLM_PROGRAM)?;
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -7,11 +7,11 @@
 
 use crate::nlm::cookie::Cookie;
 use crate::serializer::variant;
+use num_traits::ToPrimitive;
 use std::io;
 use std::io::{ErrorKind, IoSlice, Write};
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
-use num_traits::ToPrimitive;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use crate::allocator::Buffer;
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -247,7 +247,8 @@ impl<B: Buffer, T: AsyncWrite + Unpin> Serializer<B, T> {
         auth(&mut self.buffer, verifier)?;
         u64(&mut self.buffer, cookie.raw())?;
         variant(&mut self.buffer, Nlm4Stats::Granted)?;
-        Ok(self.buffer.send_inner_buffer().await?)
+        self.buffer.send_inner_buffer().await?;
+        Ok(())
     }
 
     /// Serializes [`ProcReply`] into a complete XDR RPC reply and writes it to the underlying writer.
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -11,6 +11,7 @@ use std::io;
 use std::io::{ErrorKind, IoSlice, Write};
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
+use num_traits::ToPrimitive;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use crate::allocator::Buffer;
@@ -239,7 +240,12 @@ impl<B: Buffer, T: AsyncWrite + Unpin> Serializer<B, T> {
         cred: OpaqueAuth,
         verifier: OpaqueAuth,
     ) -> io::Result<()> {
-        u32(&mut self.buffer, SystemTime::now().duration_since(UNIX_EPOCH))?;
+        let random_number_u64 = match SystemTime::now().duration_since(UNIX_EPOCH) {
+            Ok(current_time) => current_time.as_secs(),
+            Err(_) => Err(io::Error::other("SystemTime before UNIX EPOCH"))?,
+        };
+
+        u32(&mut self.buffer, (random_number_u64 >> 32) as u32)?;
         u32(&mut self.buffer, RpcBody::Call as u32)?;
         u32(&mut self.buffer, RPC_VERSION)?;
         u32(&mut self.buffer, NLM_PROGRAM)?;
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -5,17 +5,19 @@
 //! mount serializers from `crate::serializer::mount`), then emitting a complete
 //! RPC reply to an async writer.
 
+use crate::serializer::variant;
+use crate::nlm::cookie::Cookie;
 use std::io;
 use std::io::{ErrorKind, IoSlice, Write};
-
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use crate::allocator::Buffer;
+use crate::consts::nlm::{NLMPROC4_GRANTED, NLM_PROGRAM, NLM_VERSION};
 use crate::mount::MountRes;
-use crate::nlm::NlmRes;
+use crate::nlm::{Nlm4Stats, NlmRes};
 use crate::rpc::{AcceptStat, Error, OpaqueAuth, RejectedReply, ReplyBody, RpcBody};
 
-use crate::serializer::{u32, usize_as_u32, ALIGNMENT};
+use crate::serializer::{u32, u64, usize_as_u32, ALIGNMENT};
 use crate::task::{ProcReply, ProcResult};
 use crate::vfs::{NfsRes, STATUS_OK};
 
@@ -44,6 +46,9 @@ const HEADER_MASK: usize = 0x8000_0000;
 /// Size of RMS header
 /// (<https://datatracker.ietf.org/doc/html/rfc5531#autoid-19>)
 const HEADER_SIZE: usize = 4;
+
+/// Remote Procedure Call Protocol Version 2
+const RPC_VERSION: u32 = 2;
 
 macro_rules! nfs_result {
     ($self:expr, $res:expr, $ok_fn:path, $fail_fn:path) => {{
@@ -217,6 +222,32 @@ impl<B: Buffer, T: AsyncWrite + Unpin> Serializer<B, T> {
                 self.buffer.send_inner_buffer().await
             }
         }
+    }
+
+    /// Serializes a [`Cookie`] into a complete XDR RPC call and writes it to the underlying writer.
+    ///
+    /// ## Arguments:
+    /// *   `cookie` - the cookie from the original blocking LOCK request
+    /// *   `cred` - an authentication credential of [`OpaqueAuth`] type that identifies the caller to the server
+    /// *   `verifier` - an authentication verifier of [`OpaqueAuth`] type that the server generates in
+    ///     in order to validate itself to the client
+    pub async fn form_call(
+        &mut self,
+        cookie: Cookie,
+        cred: OpaqueAuth,
+        verifier: OpaqueAuth,
+    ) -> io::Result<()> {
+        u32(&mut self.buffer, rand::random::<u32>())?;
+        u32(&mut self.buffer, RpcBody::Call as u32)?;
+        u32(&mut self.buffer, RPC_VERSION)?;
+        u32(&mut self.buffer, NLM_PROGRAM)?;
+        u32(&mut self.buffer, NLM_VERSION)?;
+        u32(&mut self.buffer, NLMPROC4_GRANTED)?;
+        auth(&mut self.buffer, cred)?;
+        auth(&mut self.buffer, verifier)?;
+        u64(&mut self.buffer, cookie.raw())?;
+        variant(&mut self.buffer, Nlm4Stats::Granted)?;
+        Ok(self.buffer.send_inner_buffer().await?)
     }
 
     /// Serializes [`ProcReply`] into a complete XDR RPC reply and writes it to the underlying writer.
@@ -426,5 +457,48 @@ impl<B: Buffer, T: AsyncWrite + Unpin> WriteBuffer<B, T> {
 
         self.clean();
         Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::allocator::Slice;
+    use crate::nlm::cookie::Cookie;
+    use crate::rpc::{AuthFlavor, OpaqueAuth};
+    use std::io::Cursor;
+
+    #[tokio::test]
+    async fn form_call_serializes_rpc_call_frame() {
+        let mut buf = Cursor::new(vec![0u8; 256]);
+        let mut serializer = Serializer::<Slice, _>::new(&mut buf);
+
+        let cookie = Cookie::new(0x0A0B0C0D0E0F1011);
+        let cred = OpaqueAuth { flavor: AuthFlavor::None, body: vec![] };
+        let verifier = OpaqueAuth { flavor: AuthFlavor::None, body: vec![] };
+
+        serializer.form_call(cookie, cred, verifier).await.unwrap();
+
+        let data = buf.into_inner();
+        let body = &data[4..]; // skip RMS header
+
+        // msg_type = 0 (Call)
+        assert_eq!(body[4..8], [0, 0, 0, 0]);
+        // rpc_version = 2
+        assert_eq!(body[8..12], [0, 0, 0, 2]);
+        // program = 100021 (0x000186B5)
+        assert_eq!(body[12..16], [0, 1, 0x86, 0xB5]);
+        // version = 4
+        assert_eq!(body[16..20], [0, 0, 0, 4]);
+        // procedure = 5
+        assert_eq!(body[20..24], [0, 0, 0, 5]);
+        // cred (flavor=0, len=0)
+        assert_eq!(body[24..32], [0, 0, 0, 0, 0, 0, 0, 0]);
+        // verifier (flavor=0, len=0)
+        assert_eq!(body[32..40], [0, 0, 0, 0, 0, 0, 0, 0]);
+        // cookie = 0x0A0B0C0D0E0F1011
+        assert_eq!(body[40..48], [0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F, 0x10, 0x11]);
+        // stat = Granted = 0
+        assert_eq!(body[48..52], [0, 0, 0, 0]);
     }
 }
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -5,8 +5,8 @@
 //! mount serializers from `crate::serializer::mount`), then emitting a complete
 //! RPC reply to an async writer.
 
-use crate::serializer::variant;
 use crate::nlm::cookie::Cookie;
+use crate::serializer::variant;
 use std::io;
 use std::io::{ErrorKind, IoSlice, Write};
 use tokio::io::{AsyncWrite, AsyncWriteExt};
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -9,6 +9,8 @@ use crate::nlm::cookie::Cookie;
 use crate::serializer::variant;
 use std::io;
 use std::io::{ErrorKind, IoSlice, Write};
+use std::time::SystemTime;
+use std::time::UNIX_EPOCH;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 
 use crate::allocator::Buffer;
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -7,7 +7,6 @@
 
 use crate::nlm::cookie::Cookie;
 use crate::serializer::variant;
-use num_traits::ToPrimitive;
 use std::io;
 use std::io::{ErrorKind, IoSlice, Write};
 use std::time::SystemTime;
--- a/nfs-mamont/src/serializer/server/serialize_struct.rs
+++ b/nfs-mamont/src/serializer/server/serialize_struct.rs
@@ -471,7 +471,7 @@ mod tests {
 
     #[tokio::test]
     async fn form_call_serializes_rpc_call_frame() {
-        let mut buf = Cursor::new(vec![0u8; 256]);
+        let mut buf = Cursor::new(Vec::new());
         let mut serializer = Serializer::<Slice, _>::new(&mut buf);
 
         let cookie = Cookie::new(0x0A0B0C0D0E0F1011);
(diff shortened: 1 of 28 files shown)


nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -123,6 +123,7 @@ impl PendingLock {
     /// Returns [`Error`] if:
     /// - `caller_name` is empty.
     /// - `caller_name` is longer than [`LM_MAXSTRLEN`](nlm::LM_MAXSTRLEN).
+    #[allow(clippy::too_many_arguments)]
```
ArtemNik1tin: Is this acceptable here and [here](https://github.com/RMamonts/nfs-mamont/pull/286/changes#diff-7ea155e2572c20d8751d48e055ada6d169ce0bb990df5befc99c8a6e90db8f49R54), or do I need to figure out how to reduce the number of arguments?

ArtemNik1tin:
@VadimYarovoy Could you watch this pr? I also don't fully understand why CI is not successful, have you ever encountered such a problem?

nfs-mamont/src/serializer/server/serialize_struct.rs
```diff
@@ -219,6 +224,33 @@ impl<B: Buffer, T: AsyncWrite + Unpin> Serializer<B, T> {
...
+    /// *   `cred` - an authentication credential of [`OpaqueAuth`] type that identifies the caller to the server
+    /// *   `verifier` - an authentication verifier of [`OpaqueAuth`] type that the server generates in
+    ///     in order to validate itself to the client
+    pub async fn form_call(
+        &mut self,
+        cookie: Cookie,
+        cred: OpaqueAuth,
+        verifier: OpaqueAuth,
+    ) -> io::Result<()> {
+        u32(&mut self.buffer, rand::random::<u32>())?;
```
VadimYarovoy: not sure that rand is a good/best choice for unique XID

nfs-mamont/src/task/connection/write.rs
```diff
@@ -38,24 +42,42 @@ impl<B: Buffer> WriteTask<B> {
...
+        let granted_rx = self.granted_rx;
         let mut serializer =
             serializer::server::serialize_struct::Serializer::<B, _>::new(self.writehalf);
 
-        while let Ok(reply) = result_receiver.recv().await {
-            // TODO: <https://github.com/RMamonts/nfs-mamont/issues/143>
-            // Use proper authentication verifier instead of None
-            let verifier = OpaqueAuth { flavor: AuthFlavor::None, body: vec![] };
+        loop {
+            tokio::select! {
```
VadimYarovoy, line 50: Here are tow Ok(..) branches no else branch

if it be error then macro will panic

VadimYarovoy, line 50: one more thing

On connection teardown both senders are eventually dropped, so both recv() calls resolve to Err(RecvError), both patterns fail

can be fixed by

```rs
loop {
    tokio::select! {
        reply = result_receiver.recv() => {
            let Ok(reply) = reply else { break }; // client gone
            // ...form_reply...
        }
        cookie = granted_rx.recv() => {
            let Ok(cookie) = cookie else { continue }; // callbacks done; keep serving replies
            // ...form_call...
        }
    }
}

```

VadimYarovoy (commented):
biggest cases, will look later for smaller once

nfs-mamont/src/serializer/server/serialize_struct.rs
ArtemNik1tin: If ganesha does not work in legacy mode, then a random 32-bit number is used - https://github.com/nfs-ganesha/ntirpc/blob/96e980def1c2d4538ff4708b0908670dd6a8946d/src/rpc_com.h#L62
Also, an increment is used for each next client - https://github.com/nfs-ganesha/ntirpc/blob/96e980def1c2d4538ff4708b0908670dd6a8946d/src/clnt_generic.c#L532

nfs-mamont/src/task/connection/mod.rs
```diff
@@ -44,16 +46,17 @@ pub async fn new<A, V, B>(
...
-    )
-    .spawn();
+        result_sender.clone(),
+        granted_tx,
+    );
+    read::ReadTask::<A, B>::new(readhalf, peer_addr, context.get_write_allocator(), main_senders)
+        .spawn();
 
-    write::WriteTask::<B>::new(writehalf, result_receiver).spawn();
+    write::WriteTask::<B>::new(writehalf, result_receiver, granted_rx).spawn();
```
VadimYarovoy, line 61: they are both recievers

- result_receiver
- granted_rx

use same style pls *_receiver

nfs-mamont/src/task/connection/read.rs
```diff
@@ -22,6 +23,26 @@ use crate::task::global::nlm::NlmCommand;
 use crate::task::{ProcReply, ProcResult};
 use crate::vfs::NfsRes;
 
+pub struct MainSenders<B: Buffer + 'static> {
```
VadimYarovoy: maybe CommandSenders instead of MainSenders? 

OR if your heve better ideas - them suggest

nfs-mamont/src/task/connection/write.rs
```diff
@@ -12,6 +13,8 @@ use crate::task::ProcReply;
 pub struct WriteTask<B: Buffer> {
     writehalf: OwnedWriteHalf,
     result_receiver: async_channel::Receiver<ProcReply<B>>,
+    // The channel for sending the callback
```
VadimYarovoy: if it is doc then it should be /// insted of //

nfs-mamont/src/task/connection/write.rs
```diff
@@ -38,24 +42,44 @@ impl<B: Buffer> WriteTask<B> {
...
+                reply = result_receiver.recv() => {
+                    let Ok(reply) = reply else { break }; // client gone
+                    // TODO: <https://github.com/RMamonts/nfs-mamont/issues/143>
+                    // Use proper authentication verifier instead of None
+                    let verifier = OpaqueAuth { flavor: AuthFlavor::None, body: vec![] };
 
-            match serializer.form_reply(reply, verifier).await {
-                Ok(_) => {
-                    // Reply successfully written to socket
+                    match serializer.form_reply(reply, verifier).await {
```
VadimYarovoy: if ok branch is useless then 
```
if let Err(_) = ...
```

is more sutable

nfs-mamont/src/task/connection/write.rs
```diff
@@ -38,24 +42,44 @@ impl<B: Buffer> WriteTask<B> {
...
-                Err(e) => {
-                    error!(error=%e, "write task: failed to serialize/send reply");
-                    // TODO: Consider closing connection or continuing based on error type
-                    // For now, continue processing other replies
+                cookie = granted_rx.recv() => {
+                    let Ok(cookie) = cookie else { continue }; // callbacks done; keep serving replies
+                    let verifier = OpaqueAuth { flavor: AuthFlavor::None, body: vec![] };
+                    let credential = OpaqueAuth { flavor: AuthFlavor::None, body: vec![] };
+
+                    match serializer.form_call(cookie, credential, verifier).await {
```
VadimYarovoy: same https://github.com/RMamonts/nfs-mamont/pull/286/changes#r3691265379

nfs-mamont/src/serializer/server/serialize_struct.rs
```diff
@@ -219,6 +224,33 @@ impl<B: Buffer, T: AsyncWrite + Unpin> Serializer<B, T> {
...
+        u32(&mut self.buffer, RpcBody::Call as u32)?;
+        u32(&mut self.buffer, RPC_VERSION)?;
+        u32(&mut self.buffer, NLM_PROGRAM)?;
+        u32(&mut self.buffer, NLM_VERSION)?;
+        u32(&mut self.buffer, NLMPROC4_GRANTED)?;
+        auth(&mut self.buffer, cred)?;
+        auth(&mut self.buffer, verifier)?;
+        u64(&mut self.buffer, cookie.raw())?;
+        variant(&mut self.buffer, Nlm4Stats::Granted)?;
```
VadimYarovoy, line 256: NLMPROC4_GRANTED(nlm4_testargs) so ther is a problem in line 
```
u64(&mut self.buffer, cookie.raw())?; 
```

nfs-mamont/src/task/connection/read.rs
```diff
@@ -22,6 +23,26 @@ use crate::task::global::nlm::NlmCommand;
...
+    mount_sender: Sender<MountCommand<B>>,
+    nlm_sender: Sender<NlmCommand<B>>,
+    pool_sender: Sender<(NfsArgWrapper<B>, Sender<ProcReply<B>>)>,
+    result_sender: Sender<ProcReply<B>>,
+    granted_tx: Sender<Cookie>,
```
VadimYarovoy, line 39: according to https://github.com/RMamonts/nfs-mamont/pull/286/changes#r3691477976

here should be Sender<Nlm4TestArgs>? or all NLM args as enum?

nfs-mamont/src/serializer/server/serialize_struct.rs
VadimYarovoy: lets make it incremental too

VadimYarovoy: but make it a random on proccess start - use
```
SystemTime::now().duration_since(UNIX_EPOCH)
```

instead of rand for this randomization

nfs-mamont/src/task/connection/write.rs
VadimYarovoy: find out that
> An NLM callback is a separate RPC call from the server to the client. The client (Linux lockd) starts its own NLM service and registers it with rpcbind, while the server must establish a new connection to the client's lockd port and issue a CALL. This is precisely how `nlmsvc_grant_callback` works in the Linux kernel.

so reply should be send in another connection - look how client works, probably on server start here must be one more tcp connection with client

nfs-mamont/src/serializer/server/serialize_struct.rs
VadimYarovoy: The XID is generated in `form_call` and immediately lost. The client response (`nlm4_res` with `stat`) is not awaited anywhere

GRANTED -- a call with unreliable semantics: the client might have rebooted, lost state, or returned NLM4_DENIED (the process waiting for the lock was killed). The specification implies that the server retries GRANTED until a response is received, and upon receiving DENIED, releases the granted lock; otherwise, the lock remains held indefinitely by a dead owner.

Current behavior: a single attempt, result unknown. If the packet is lost or the client reboots, the lock remains "hanging" until the server restarts.

VadimYarovoy: if is not possible to fix in that PR - write why, then TODO will be left in code for futher development

nfs-mamont/src/service/nlm/tests/operations/unlock.rs
```diff
@@ -37,26 +44,81 @@ async fn unlock_preserves_cookie() {
...
+    assert_eq!(blocked.stat, Nlm4Stats::Blocked);
+
+    svc.unlock(make_unlock_args(FH_DEFAULT, "alice", 100, 1)).await;
+
+    let received = rx.recv().await.expect("channel should not be closed");
+    assert_eq!(received.raw(), 99);
+}
+
+#[tokio::test]
+async fn unlock_no_callback_when_granted_tx_is_none() {
```
VadimYarovoy, line 98: seems like there is no callback at all (even no channel for callback) so this test does not test that there is no callback

VadimYarovoy: need to be channel in this test to ckeck that no call back is present

```
#[tokio::test]
async fn unlock_no_callback_when_granted_tx_is_none() {
    let svc = NlmService::new();
    let (tx, rx) = async_channel::unbounded::<Cookie>();

    svc.lock(None, make_lock_args_without_block(FH_DEFAULT, true, 0, LOCK_WHOLE_LENGTH, "alice", 100, 0)).await;
    let blocked = svc
        .lock(None, make_lock_args_with_block(FH_DEFAULT, true, 0, LOCK_WHOLE_LENGTH, "bob", 200, 99))
        .await;
    assert_eq!(blocked.stat, Nlm4Stats::Blocked);

    svc.unlock(make_unlock_args(FH_DEFAULT, "alice", 100, 1)).await;

    assert!(matches!(rx.try_recv(), Err(async_channel::TryRecvError::Empty)));
    drop(tx);
}
```

nfs-mamont/src/serializer/server/serialize_struct.rs
```diff
@@ -428,3 +460,46 @@ impl<B: Buffer, T: AsyncWrite + Unpin> WriteBuffer<B, T> {
...
+mod tests {
+    use super::*;
+    use crate::allocator::Slice;
+    use crate::nlm::cookie::Cookie;
+    use crate::rpc::{AuthFlavor, OpaqueAuth};
+    use std::io::Cursor;
+
+    #[tokio::test]
+    async fn form_call_serializes_rpc_call_frame() {
+        let mut buf = Cursor::new(vec![0u8; 256]);
```
VadimYarovoy: A `Cursor` over a `vec![0u8; 256]` overwrites bytes rather than appending to them: the vector's length remains 256, and any bytes untouched by the serializer remain zeros. Consequently, any check like `assert_eq!(..., [0,0,0,0])` is effectively meaningless—including the check for `stat`, which cannot distinguish between a written zero and nothing having been written at all. If the `stat` field were omitted from `form_call` entirely, the test would still pass.

VadimYarovoy: fix in next way

```
let mut buf = Cursor::new(Vec::new());   // no zeroes
...
let data = buf.into_inner();

// RM-header: last fragment + body len
let size = (data.len() - 4) as u32;
assert_eq!(u32::from_be_bytes(data[0..4].try_into().unwrap()), 0x8000_0000 | size);

// check frame len
assert_eq!(data.len(), 4 /*RM*/ + 4 /*xid*/ + 20 /*call header*/ + 16 /*auth*/ + <len(args)>);
```

nfs-mamont/src/service/nlm/unlock.rs
```diff
@@ -24,9 +24,18 @@ impl Unlock for NlmService {
...
-            return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed };
+
+        let granted = match registry.grant_pending(&fh) {
+            Ok(granted) => granted,
+            Err(_) => return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed },
+        };
+
+        for lock in granted {
+            if let Some(tx) = lock.grant_notification.granted_tx {
+                if let Err(e) = tx.send(lock.grant_notification.cookie).await {
```
VadimYarovoy, line 38: await under guard

this guard
```
let mut registry = self.locks.write().await;
```

if this will become bounded then dealock is possible. 
Even without a deadlock, this constitutes an unnecessary hold on the global lock during an I/O-like operation.

can be fixed by

```rs
let granted = {
    let mut registry = self.locks.write().await;
    if registry.remove_by_owner(...).is_err() {
        return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed };
    }
    match registry.grant_pending(&fh) {
        Ok(granted) => granted,
        Err(_) => return Nlm4UnlockRes { cookie: args.cookie, stat: Nlm4Stats::Failed },
    }
};   // guard droped

for lock in granted {
    if let Some(tx) = &lock.grant_notification.granted_tx {
        if let Err(e) = tx.send(...).await {
            tracing::warn!(error = %e, "failed to send grant callback");
        }
    }
}
```

nfs-mamont/src/service/nlm/mod.rs
```diff
@@ -24,6 +24,21 @@ mod unlock;
 #[cfg(test)]
 mod tests;
 
+/// The wrapper needed to notify the client.
+struct GrantNotification {
+    /// Identifies a point in the directory.
```
VadimYarovoy: incorrect comment. this field about transaction

VadimYarovoy: /// Transaction identifier from the original blocking LOCK request;
/// echoed back to the client in the GRANTED callback.

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -1,10 +1,12 @@
...
 use crate::nlm::procedures::lock::{Lock, Nlm4LockArgs, Nlm4LockRes};
 use crate::nlm::Nlm4Stats;
+use async_channel::Sender;
 
-use super::{ActiveLock, NlmService, PendingLock};
+use super::{ActiveLock, GrantNotification, NlmService, PendingLock};
 
 impl Lock for NlmService {
-    async fn lock(&self, args: Nlm4LockArgs) -> Nlm4LockRes {
+    async fn lock(&self, granted_tx: Option<Sender<Cookie>>, args: Nlm4LockArgs) -> Nlm4LockRes {
```
VadimYarovoy, line 9:
```
granted_tx: Option<Sender<Cookie>>
```

looks not good here - maybe it is better to move it inside Nlm4LockArgs and not to parse/serialize?

nfs-mamont/src/task/connection/read.rs
VadimYarovoy: moreover add doc to struct and it fields (because it is pub)

nfs-mamont/src/task/connection/write.rs
```diff
@@ -38,24 +42,44 @@ impl<B: Buffer> WriteTask<B> {
...
+                        Ok(_) => {
+                            // Call successfully written to socket
+                        }
+                        Err(e) => {
+                            error!(error=%e, "write task: failed to serialize/send reply");
```
VadimYarovoy: err log need specification - it is same as in branch above

nfs-mamont/src/service/nlm/unlock.rs
VadimYarovoy: lets make this channel bounded - i will make other channels bounded too later (backpressure mechanism)

VadimYarovoy (commented):
find a lot of things to think about

nfs-mamont/src/service/nlm/lock.rs
ArtemNik1tin: I can do it, but it contradicts the RFC(https://pubs.opengroup.org/onlinepubs/9629799/chap10.htm#tagcjh_11_02_02_09).

nfs-mamont/src/service/nlm/unlock.rs
ArtemNik1tin: What channel capacity should be used?

nfs-mamont/src/service/nlm/lock.rs
MikhailSvirukov: Probably even though we does not give that trait to user, with that recv we does tie interface to inner logic of server - which does not look good. But, somehow we need to associate pending lock with Writer task it should go once acquired. So, probably we could solve that by holding clone of Sender inside NLM state. But, there is a problem of writer task identification, that we does not have now. So, when there is new connection created, it does append itself to nlm state (even if it's rwlock - probably registering is the only one that would require unique access). And probably, since there are more procedures in nlm it could be simplier. @VadimYarovoy, what do you think?

ElementaryBelkin:
> Per RFC 1813, when a server responds with `NLM4_BLOCKED` to a blocking lock request, it must send an asynchronous `NLMPROC4_GRANTED` callback RPC once the lock becomes available. Currently, `grant_pending()` returns the newly-granted locks but the caller discards the result.
> 
> This PR adds:
> 
> * RPC callback channel (`granted_tx`/`granted_rx`) per connection for server-initiated `NLMPROC4_GRANTED`
> * Cookie delivery via channel in `unlock.rs` after `grant_pending()`
> * RPC CALL serialization in `Serializer::form_call()`
> * `tokio::select!` in `WriteTask` for multiplexing replies and callbacks
> * Tests for serialization, callback flow, and no-callback path
> 
> Resolves #267

А по-английски то без переводчика пишешь?

ArtemNik1tin:
@VadimYarovoy @MikhailSvirukov 
I’ve come up with an idea for how we can keep the NLM trait clean. What if a person who wants to change the implementation of some method sends an event to the `mpsc::UnboundedSender<NlmEvent>` channel? And inside NlmTask, we’ll create two handlers: `handle_command()` and `handle_event()`. So far, I can see only two options for how we can pass the channel for events:
1. We’ll pass it in the constructor when creating NlmService.
2. Each nlm method will return a pair (result, events)
```
pub struct NlmOutcome<T> {
    pub response: T,
    pub events: Vec<NlmEvent>,
}
```

I think the second option is more convenient for the client, but then we’re strictly tied to the fact that events occur exclusively during procedures and not in the background

nfs-mamont/src/serializer/server/serialize_struct.rs
```diff
@@ -35,6 +39,21 @@ use super::rpc::auth;
 /// this size is enough to hold only arguments without opaque data ([`Buffer`] in [`crate::vfs::read::Success`])
 const DEFAULT_SIZE: usize = 4096;
 
+/// Max size of RMS fragment data
+/// (<https://datatracker.ietf.org/doc/html/rfc5531#autoid-19>)
+const MAX_FRAGMENT_SIZE: usize = 0x7FFF_FFFF;
```
MikhailSvirukov: Duplicates from `consts/rpc`

nfs-mamont/src/nlm/mod.rs
```diff
@@ -54,16 +55,22 @@ pub enum Nlm4Stats {
...
-    /// NLM TEST procedure response.
+    /// Server response to the NLM TEST procedure.
     Test(Box<Nlm4TestRes>),
-    /// NLM CANCEL procedure response.
+    /// Server response to the NLM CANCEL procedure.
     Cancel(Nlm4CancelRes),
 }
 
+/// Wrapper for all supported types of procedures that call the client.
+pub enum NlmCall {
```
MikhailSvirukov: Maybe `NlmCallbackReply`?

nfs-mamont/src/serializer/server/nlm/mod.rs
```diff
@@ -3,14 +3,16 @@
...
+use crate::nlm::procedures::test::Nlm4TestArgs;
 use crate::nlm::procedures::{
     cancel::Nlm4CancelRes, lock::Nlm4LockRes, test::Nlm4TestRes, unlock::Nlm4UnlockRes,
 };
-use crate::nlm::Nlm4Stats;
-use crate::serializer::{u32, u64, variant, vector};
+use crate::nlm::{Nlm4Stats, OpaqueHandle};
+use crate::serializer::files::file_handle;
+use crate::serializer::{bool, i32, string, u32, u64, variant, vector};
+use std::io;
```
MikhailSvirukov: import order

nfs-mamont/src/nlm/mod.rs
```diff
@@ -54,16 +55,22 @@ pub enum Nlm4Stats {
...
+    /// Client response to the NLM GRANTED procedure.
+    Granted(Nlm4TestArgs),
```
MikhailSvirukov: And why Granted has `Nlm4TestArgs` type?

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -14,7 +19,7 @@ impl Lock for NlmService {
             args.lock.lock_offset,
             args.lock.lock_length,
             args.lock.opaque_handle,
-            args.cookie,
+            PendingGrant::new(rpc_call_sender.clone(), args.cookie),
```
MikhailSvirukov: remove clone

nfs-mamont/src/service/nlm/lock.rs
```diff
@@ -1,10 +1,15 @@
...
-
-use super::{ActiveLock, NlmService, PendingLock};
+use crate::task::ProcCall;
+use async_channel::Sender;
 
 impl Lock for NlmService {
-    async fn lock(&self, args: Nlm4LockArgs) -> Nlm4LockRes {
+    async fn lock(
+        &self,
+        rpc_call_sender: Option<Sender<ProcCall>>,
```
MikhailSvirukov: @ArtemNik1tin could you, please show in this pr (just for simplicity), how that would be flexible in terms of one's outer implementations? - i just do not get it. I do get it, that that channel goes way inside outer lock logic and ones acquired outer implementation to send reply by this channel. But is it? And how does that correlate with the fact, that our implementation calls it inside Unlock?

Also, @VadimYarovoy and @ArtemNik1tin - how are we planning to use Nlm trait? I mean, we did implement (actually not, but let's thing that we did) some sync methods. What's next? There are more methods in specification, are we planning to add them to trait, or somehow implement with things we have, or something third? I have feeling, that some things might be impossible to implement  (or make abstraction flexible enough) without information about out inner architecture. I mean, maybe we ought to think, where we going with Nlm trait and thing about it globally, since we have so much trouble with single method extension.

nfs-mamont/src/nlm/mod.rs
ArtemNik1tin: That’s what the specification says

ArtemNik1tin: https://pubs.opengroup.org/onlinepubs/9629799/NLM_GRANTED.htm#tagcjh_11_08

nfs-mamont/src/service/nlm/lock.rs
ArtemNik1tin: https://github.com/RMamonts/nfs-mamont/discussions/321

nfs-mamont/src/task/connection/write.rs
ArtemNik1tin: Am I correct in understanding that this issue can be considered resolved?

ArtemNik1tin:
@VadimYarovoy I propose to put the implementation of granted in the next pr

nfs-mamont/src/task/connection/mod.rs
```diff
@@ -44,16 +45,18 @@ pub async fn new<A, V, B>(
     let (result_sender, result_receiver) = async_channel::unbounded::<ProcReply<B>>();
```
ArtemNik1tin: I plan to combine this channel with ProcCall if ProcCall is not needed for asynchronous procedures

PR #288, opened 2026-07-30
fix(CI): force JavaScript actions to use Node 24

ArtemNik1tin:
GitHub Actions has started deploying Node 24 on runners by default, Node 20 is marked as deprecated. JavaScript actions can continue to use Node 20 without explicit instructions, which leads to warnings:

> Node 20 is being deprecated. This workflow is running with Node 24 by default.

Added the environment variable `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` to both workflow files (`.github/workflows/rust-ci.yml` and `.github/workflows/fuzz.yml`), which forces the launch of all JavaScript actions on Node 24.


.github/workflows/rust-ci.yml
```diff
@@ -83,15 +84,21 @@ jobs:
...
         uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
 
-      - name: Run cargo-audit
-        uses: rustsec/audit-check@v2
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install cargo-audit
+        run: cargo install cargo-audit --locked
```
VadimYarovoy: here is a problem because `cargo install cargo-audit --locked` compiles cargo-audit and all its dependencies from source on every CI run

VadimYarovoy: need some kind of prebuild instance

like 
```
      - name: Install cargo-audit
        uses: taiki-e/install-action@v2
```

may be there are others

.github/workflows/rust-ci.yml
```diff
@@ -83,15 +84,21 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
+      - name: Install Rust toolchain
```
VadimYarovoy: remove it too if `cargo install cargo-audit --locked` will be replaced with prebuilded action

.github/workflows/rust-ci.yml
VadimYarovoy: or even use old one github action and just update rust max version to 1.96.0, because problem was
```
error: failed to compile `cargo-audit v0.22.2`, intermediate artifacts can be found at `/tmp/cargo-installJxUg05`.
  To reuse those artifacts with a future compilation, set the environment variable `CARGO_BUILD_BUILD_DIR` to that path.
  
  Caused by:
    rustc 1.95.0 is not supported by the following package:
      kstring@2.0.4 requires rustc 1.96.0
    Try re-running `cargo install` with `--locked`
```

VadimYarovoy (approved):

PR #289, opened 2026-08-11
refactor(allocator): unify buffer pools and move read alloc to parser

VadimYarovoy:
Remopved read allocator. All allocations now are inside readtask

### DIFF
--- a/nfs-mamont/src/parser/parser_struct.rs
+++ b/nfs-mamont/src/parser/parser_struct.rs
@@ -257,7 +257,10 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
             READLINK => {
                 NfsArguments::ReadLink(self.buffer.parse_with_retry(read_link::args).await?)
             }
-            READ => NfsArguments::Read(self.buffer.parse_with_retry(read::args).await?),
+            READ => {
+                let (args, data) = adapter_for_read(&self.allocator, &mut self.buffer).await?;
+                NfsArguments::Read(args, data)
+            }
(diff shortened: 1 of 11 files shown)


nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -646,6 +649,46 @@ where
...
+    S: AsyncRead + Unpin,
+{
+    let args = buffer.parse_with_retry(read::args).await?;
+
+    let data = if args.count == 0 {
+        A::Buffer::empty()
+    } else {
+        let size = NonZeroUsize::new(args.count as usize).unwrap();
+        alloc.allocate(size).await.ok_or_else(|| {
+            Error::IO(io::Error::new(ErrorKind::OutOfMemory, "cannot allocate memory"))
```
MikhailSvirukov, line 704: Actually, good point from copilot (though nfs error `TooSmall` looks wrong too) - maybe returning nfs level error like `NFS3ERR_SERVERFAULT`. Although, there is not appropriate code as i see it in protocol description

MikhailSvirukov: Maybe it is good thing, that our structure does not allow explicit nfs errors here)

MikhailSvirukov (dismissed):

nfs-mamont/src/parser/parser_struct.rs
VadimYarovoy: now requests are shrinked to max allocator  size

MikhailSvirukov (approved):

PR #290, opened 2026-08-12
allocation: make drop on each list item rather than whole list

MikhailSvirukov:
Currently resource managing happen on level of list - list items return only on whole list drop. That might be a problem in case backend might use separate chunks longer than other - that might cause to whole system to wait for allocation longer, when actually some of chunks are ready for another request.

`Impl2` is an experimental example of how such approach might be used


nfs-mamont/src/allocator/buffer/dropbuf.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn is_empty(&self) -> bool {
+        self.len == 0
+    }
+}
+
+impl Drop for UnownedDroppableBuffer {
+    fn drop(&mut self) {
+        let clone = unsafe {
+            UnownedDroppableBuffer::from_raw_parts(self.ptr, self.len, self.state.clone())
```
MikhailSvirukov: it does look creepy, but actually looks like it is quite cheap

MikhailSvirukov: Actually, the idea is to have original deallocator from some other type buffer we conver this one. So we shouldn't do drop, because it is intended to be called from somewhere else

nfs-mamont/src/allocator/impl2.rs
```diff
@@ -0,0 +1,100 @@
...
+        let mut buffers = Vec::with_capacity(count_needed);
+        for _ in 0..count_needed {
+            if let Some(buf) = self.state.pool.pop() {
+                buffers.push(buf);
+            } else {
+                unreachable!("Semaphore permitted allocation but pool was empty");
+            }
+        }
+
+        permit.forget();
```
VadimYarovoy, line 92: here is forget (looks stange in `alloc()` method but OK)

futhermore  impl Drop у SliceNoDrop and UnownedDroppableBuffer were deleted in commit a8c05ba

nfs-mamont/src/allocator/impl2.rs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            let ptr = base_ptr as *mut libc::c_void;
+            if unsafe { libc::mlock(ptr, total_size) } != 0 {
+                let err = io::Error::last_os_error();
+                panic!("mlock failed (size={}): {err}", size.get());
+            }
+        }
+
+        let mut current_ptr = base_ptr;
+        let state = Arc::new(AllocatorState { pool, semaphore, base_ptr, layout });
```
VadimYarovoy, line 51: maybe ARC cylce - need to look better

VadimYarovoy (commented):
May be i didn't understand idea, will look one more time later

PR #291, opened 2026-08-13
refactor(allocator): introduce non-blocking try_allocate with fallback

VadimYarovoy:
Replace blocking Allocator::allocate with best-effort try_allocate and a separate allocate_block method. The default allocate implementation now attempts non-blocking allocation first, falling back to a single block when the pool is exhausted.

Update the RPC parser to handle partial allocations by reading only the available bytes and discarding the rest to maintain stream alignment. Eliminate OutOfMemory errors in favor of this best-effort approach.

BREAKING CHANGE: Allocator trait now requires implementing try_allocate and allocate_block. allocate is now a default method combining both.

### DIFF
--- a/nfs-mamont/src/allocator/mod.rs
+++ b/nfs-mamont/src/allocator/mod.rs
@@ -70,26 +70,53 @@ pub trait Allocator {
     /// Type of buffer returned by this allocator.
     type Buffer: Buffer;
 
-    /// Returns a buffer of at least `size` bytes.
+    /// Attempts to allocate a buffer of (at most) `size` bytes without blocking.
     ///
-    /// Waits until enough memory is available, so the returned future may stay
-    /// pending while the pool is exhausted.
+    /// This is a "best-effort" operation: the returned buffer may be smaller
+    /// than `size` (partial allocation). If no memory can be granted right now,
+    /// [`None`] is returned — the caller should either retry later or fall back
+    /// to the blocking [`Self::allocate_block`].
     ///
     /// # Parameters
     ///
-    /// - `size` --- minimum size of the returned buffer in bytes.
-    ///
-    /// # Returns
-    ///
-    /// Returns [`None`] if `size` is greater than [`Allocator::capacity`],
-    /// since such a request can never be satisfied.
-    fn allocate(&self, size: NonZeroUsize) -> impl Future<Output = Option<Self::Buffer>> + Send;
+    /// - `size` --- maximum size of the returned buffer in bytes.
+    fn try_allocate(&self, size: NonZeroUsize) -> Option<Self::Buffer>;
 
-    /// Returns the largest size [`Allocator::allocate`] can ever satisfy.
+    /// Allocates a single minimal buffer, blocking until one is available.
+    ///
+    /// The returned buffer is exactly one allocation unit of the allocator and
+    /// is not guaranteed to satisfy the originally requested `size`.
+    fn allocate_block(&self) -> impl Future<Output = Self::Buffer> + Send;
+
+    /// Returns a buffer of at most `size` bytes.
+    ///
+    /// This is a convenience combination of [`Self::try_allocate`] and
+    /// [`Self::allocate_block`]: it first attempts a non-blocking allocation,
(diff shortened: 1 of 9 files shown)


nfs-mamont/src/allocator/mod.rs
```diff
@@ -59,6 +59,18 @@ pub trait Buffer: Send + Sync {
     /// Returns `true` if the buffer is empty.
     fn is_empty(&self) -> bool;
 
+    /// Shrinks the buffer to at most `len` bytes.
+    ///
+    /// Buffers handed out by an allocator may be larger than the payload that
+    /// was actually written into them (see [`Allocator::allocate_block`]).
+    /// Truncating hides the untouched tail --- which still holds bytes of
```
MikhailSvirukov, line 106: Maybe remove line about untouched tail? I mean such behavior might be left to implementation - it does not really break anything

nfs-mamont/src/allocator/mod.rs
```diff
@@ -151,21 +196,32 @@ impl Impl {
...
+        let mut count = wanted.div_ceil(self.buffer_size.get());
+        let permit = loop {
+            if count == 0 {
+                return None;
+            }
+            match self.state.semaphore.try_acquire_many(count as u32) {
+                Ok(permit) => break permit,
+                Err(_) => {
+                    let available = self.state.semaphore.available_permits();
+                    count = count.saturating_sub(1).min(available);
```
MikhailSvirukov: I am not sure, that in case many request for large portions case at the same time, such approach will be good - looks like trashing, if we keep in mind, that there is also next step, that returns exactly single buffer

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -632,30 +646,58 @@ where
...
+
     // A zero-length write needs no buffer at all, so it never takes a pool slot.
+    // Otherwise attempt a non-blocking allocation and fall back to a single
+    // blocking block when the pool cannot satisfy the request right now.
     let mut buffer_data = match NonZeroUsize::new(size) {
         None => A::Buffer::empty(),
-        Some(non_zero_size) => alloc.allocate(non_zero_size).await.ok_or_else(|| {
-            Error::IO(io::Error::new(ErrorKind::OutOfMemory, "cannot allocate memory"))
-        })?,
+        Some(non_zero_size) => match alloc.try_allocate(non_zero_size) {
```
MikhailSvirukov, line 671: Maybe add method like "allocate_guaranted" and move that allocation in it?

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -632,30 +646,58 @@ where
...
+        read_in_slice_async(buffer, &mut buffer_data, bytes_read_sync, to_read - bytes_read_sync)
             .await?;
     }
 
-    // Discard any trailing padding bytes after the data.
-    buffer.discard_bytes(padding).await.map_err(Error::IO)?;
+    // A fallback block can be larger than the payload; its tail still holds the
+    // bytes of whatever request owned the block before. Cut it off so nothing
+    // downstream can write those bytes to a file.
+    buffer_data.truncate(to_read);
```
MikhailSvirukov, line 691: Also i am a bit concerned about truncate - in Seriaizer we send buffer whole, so we ignore actual size in `size` field of reply, which means there might be some data, user never intended to write (and even some data, that wasn't clean). Maybe we should make some quarantees about size, that is used in buffer, and actuall buffer size, rather than make do explicit call

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -632,30 +646,58 @@ where
...
+
+    // Discard the unread data remainder and any trailing padding bytes.
+    buffer.discard_bytes(size - to_read + padding).await.map_err(Error::IO)?;
```
MikhailSvirukov, line 694: That should definitely be tested) Overall, it looks like we do force nfs client to send the same data again and again... maybe simply waiting for allocation. Though it might be of help to overall server performance.

PR #293, opened 2026-08-19
feat: simplify parser

### LINKED ISSUE #292: parser: simplification
Basically, paser logic seems to be a bit complicated. Two buffered logic, that lies underneath might be simplified to single buffer instead of current version with two while preserving existing logic (or even enhance in some ways)

MikhailSvirukov:
This pr introduced changes to the structure of parser object.

Previous version used two-buffer logic, when parser buffers each next request in different one, that was used in previous request; also that required special (and quite complicated) logic to handle partial request receiving, when parser had to restore it's state to read missing part and try again.

New version basically implies the same logic, but uses single buffer; instead of "retrying" to parse incomplete request, it ensures, that request is fully received, and then starts to parse request. 

PR copies specific of parsing changes form `next` branch

Resolves: #292

### DIFF
--- a/nfs-mamont/src/parser/parser_struct.rs
+++ b/nfs-mamont/src/parser/parser_struct.rs
@@ -1,17 +1,20 @@
 //! RPC message parser for NFS, MOUNT and NLM protocols.
 //!
-//! This module provides the [`RpcParser`] struct, which parses XDR-encoded RPC messages
+//! This module provides [`RpcParser`] struct, which parses XDR-encoded RPC messages
 //! according to RFC 5531 (RPC) and RFC 1813 (NFSv3). It handles:
 //!
 //! - RPC message framing and headers
-//! - Authentication (currently only AUTH_NONE)
+//! - Authentication (AUTH_NONE and AUTH_SYS)
 //! - NFSv3 procedure parsing (all 22 procedures)
 //! - MOUNT protocol procedure parsing
 //! - NLM procedure parsing
 //! - Error handling and message discarding on protocol errors
 //!
-//! The parser uses a [`CountBuffer`] to efficiently read from async streams while
-//! supporting retry logic for parsing operations that may need additional data.
+//! The parser uses a [`FrameReader`] which buffers head window of every
+//! RMS frame, so headers and procedure arguments are parsed synchronously
+//! without retries. Only opaque `WRITE` payloads that do not fit into
+//! buffered window are streamed from the socket directly into the allocated
+//! [`Buffer`].
 
 use std::cmp::min;
 use std::io::{self, ErrorKind};
@@ -43,21 +46,42 @@ use crate::parser::nfsv3::{
 };
 use crate::parser::nlm::{cancel::cancel, lock::lock, test::test, unlock::unlock};
 use crate::parser::primitive::{u32, u32_as_usize, ALIGNMENT};
-use crate::parser::read_buffer::CountBuffer;
-use crate::parser::rpc::{auth, RpcMessage};
+use crate::parser::read_buffer::FrameReader;
+use crate::parser::rpc::{auth, authsys_parms, RpcMessage};
 use crate::parser::{
     proc_nested_errors, ArgWrapper, Error, ErrorWrapper, MountArgWrapper, MountArguments,
     NfsArgWrapper, NfsArguments, NlmArguments, ProcArguments, Result, RpcHeader,
 };
-use crate::rpc::{AuthFlavor, AuthStat, OpaqueAuth, RpcBody, VersionMismatch, RPC_VERSION};
+use crate::rpc::{AuthFlavor, AuthStat, Credential, RpcBody, VersionMismatch, RPC_VERSION};
 use crate::vfs;
-
-const RMS_HEADER_SIZE: usize = size_of::<u32>();
 
 /// Minimum buffer size, that could hold complete RPC message
 /// with NFSv3 or Mount protocol arguments, except for NFSv3 `WRITE` procedure -
-/// this size is enough to hold only arguments without opaque data ([`Buffer`] in [`vfs::write::Args`])
+/// this size is enough to hold only arguments without opaque data ([`Buffer`] in [`vfs::write::Args`]).
+///
+/// The arguments of every procedure except for the opaque `WRITE` data MUST fit
+/// into the parser buffer: the parser reads at most `capacity` bytes of a
+/// frame into memory before parsing and fails with `InvalidData` if the
+/// arguments extend beyond that window.
 pub const DEFAULT_SIZE: usize = 2500;
+
+/// Maps an `UnexpectedEof` from synchronous argument parsing to `InvalidData`.
+///
+/// After [`FrameReader::begin_body`] the head window of the frame is fully
+/// buffered, so running out of data during synchronous parsing means either
+/// the frame is shorter than its arguments (truncated message) or the
+/// arguments do not fit into the parser buffer (legal only for `WRITE`
+/// payloads, which are streamed separately). Both cases are protocol
+/// violations rather than transient I/O conditions.
+fn map_eof(error: Error) -> Error {
+    match error {
+        Error::IO(err) if err.kind() == ErrorKind::UnexpectedEof => Error::IO(io::Error::new(
+            ErrorKind::InvalidData,
+            "RPC message arguments exceed frame or buffer bounds",
+        )),
+        other => other,
+    }
+}
 
 /// Parser for RPC messages over async streams.
 ///
@@ -74,9 +98,7 @@ pub const DEFAULT_SIZE: usize = 2500;
 /// * `S` - An async stream type that implements [`AsyncRead`] and [`Unpin`]
 pub struct RpcParser<A: Allocator, S: AsyncRead + Unpin> {
     allocator: Arc<A>,
-    buffer: CountBuffer<S>,
-    last: bool,
-    current_frame_size: usize,
+    reader: FrameReader<S>,
 }
 
 impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
@@ -91,12 +113,7 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
     ///
     /// A new `RpcParser` instance ready to parse messages.
     pub fn new(socket: S, allocator: Arc<A>) -> Self {
-        Self {
-            allocator,
-            buffer: CountBuffer::new(DEFAULT_SIZE, socket),
-            last: false,
-            current_frame_size: 0,
-        }
+        Self { allocator, reader: FrameReader::new(DEFAULT_SIZE, socket) }
     }
 
     /// Creates a new `RpcParser` with the specified buffer size.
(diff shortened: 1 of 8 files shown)


nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -43,22 +46,43 @@ use crate::parser::nfsv3::{
...
 /// Minimum buffer size, that could hold complete RPC message
 /// with NFSv3 or Mount protocol arguments, except for NFSv3 `WRITE` procedure -
-/// this size is enough to hold only arguments without opaque data ([`Buffer`] in [`vfs::write::Args`])
+/// this size is enough to hold only arguments without opaque data ([`Buffer`] in [`vfs::write::Args`]).
+///
+/// The arguments of every procedure except for the opaque `WRITE` data MUST fit
+/// into the parser buffer: the parser reads at most `capacity` bytes of a
+/// frame into memory before parsing and fails with `InvalidData` if the
+/// arguments extend beyond that window.
 pub const DEFAULT_SIZE: usize = 2500;
```
MikhailSvirukov, line 66: @VadimYarovoy what do you think of enlarging it to be able to hold more than single request? Main concern is that it does potentially increase amount of opaque data, that would be copied from underlying buffer into Slice. Probably it wouldn't be much and probably still be fast, but maybe (?) it could help with small multiple requests

VadimYarovoy: this can be judged in case of performance only after measurements, to my mind it is impossible to say will it be faster or not

VadimYarovoy:
@MikhailSvirukov make good pr description pls for other reviewers

VadimYarovoy:
will look this PR at evening or tomorrow

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -687,123 +694,22 @@ where
...
-    args.count = count as u32;
-
-    let data = match NonZeroUsize::new(count) {
-        None => A::Buffer::empty(),
-        Some(size) => alloc.allocate(size).await.ok_or_else(|| {
+    let data = if args.count == 0 {
+        A::Buffer::empty()
+    } else {
+        let size = NonZeroUsize::new(args.count as usize).unwrap();
+        alloc.allocate(size).await.ok_or_else(|| {
```
VadimYarovoy, line 659: missed case

```
None => A::Buffer::empty()
```

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -585,73 +587,84 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
+    if size + padding > reader.frame_remaining() {
+        return Err(Error::IO(io::Error::new(
+            ErrorKind::InvalidData,
+            "WRITE data length exceeds frame size",
+        )));
+    }
+
+    // Attempt allocation with the given size, or fallback to NonZeroUsize::MIN.
+    let non_zero_size = NonZeroUsize::new(size).unwrap_or(NonZeroUsize::MIN);
+    let mut buffer_data = alloc.allocate(non_zero_size).await.ok_or_else(|| {
```
VadimYarovoy: missed case with 0 size payload

```
None => A::Buffer::empty()
```

to my mind we shouldnt give allocator blocks on such requests

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -585,73 +587,84 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
-            .await?;
+    // The opaque data with its padding must lie within the current frame;
+    // otherwise, the declared length is bogus and reading it would consume
+    // bytes of the next message, misaligning the stream.
+    if size + padding > reader.frame_remaining() {
+        return Err(Error::IO(io::Error::new(
+            ErrorKind::InvalidData,
+            "WRITE data length exceeds frame size",
+        )));
+    }
```
VadimYarovoy, line 589: overflow is possible 

can be done something like

```rs
let bounds_error = || {
    Error::IO(io::Error::new(ErrorKind::InvalidData, "WRITE data length exceeds frame size"))
};
let padded = size.checked_add(padding).ok_or_else(bounds_error)?;
if padded > reader.frame_remaining() {
    return Err(bounds_error());
}
```

maybe this is ok, idk

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -556,18 +563,13 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
     ///
     /// Returns the error, potentially after attempting to discard the message.
     async fn match_errors(&mut self, error: Error) -> Error {
```
VadimYarovoy, line 514: probably should be

```rs
async fn match_errors(&mut self, error: Error) -> Error {
    if self.reader.frame_remaining() == 0 {
        return error;
    }
    proc_nested_errors(error, self.discard_current_message()).await
}
```

because if  frame_remaining > 0, and next `read_frame_header()` reads data as  RMS header then connection will break

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -388,21 +423,18 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
             Err(err) => return Err(self.match_errors(err).await),
         };
-        let proc = match self.parse_mount_message_with_header(&rpc_header).await {
+        let proc = match self.parse_mount_message_with_header(&rpc_header) {
             Ok(arg) => Box::new(arg),
             Err(err) => return Err(self.match_errors(err).await),
         };
 
         // finalize_parsing() is only called after successful header and procedure parsing; it is not run on error paths
         self.finalize_parsing()?;
```
VadimYarovoy:
```rs
if let Err(error) = self.finalize_parsing() {
    return Err(self.match_errors(error).await);
}
```

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -344,10 +385,7 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
```
VadimYarovoy:
```rs
if let Err(error) = self.finalize_parsing() {
    return Err(self.match_errors(error).await);
}
```

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -376,10 +414,7 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
 
         // finalize_parsing() is only called after successful header and procedure parsing; it is not run on error paths
         match self.finalize_parsing() {
```
VadimYarovoy, line 383:
```rs
if let Err(error) = self.finalize_parsing() {
    return Err(ErrorWrapper { xid: Some(xid), error: self.match_errors(error).await });
}
Ok(ArgWrapper { header: RpcHeader { xid, cred: rpc_header.cred }, proc })
```

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -585,73 +587,84 @@ impl<A: Allocator, S: AsyncRead + Unpin> RpcParser<A, S> {
...
-    // A zero-length write needs no buffer at all, so it never takes a pool slot.
-    let mut buffer_data = match NonZeroUsize::new(size) {
-        None => A::Buffer::empty(),
-        Some(non_zero_size) => alloc.allocate(non_zero_size).await.ok_or_else(|| {
-            Error::IO(io::Error::new(ErrorKind::OutOfMemory, "cannot allocate memory"))
-        })?,
-    };
+    // Parse arguments for the WRITE procedure.
+    let part_arg = write::args(reader).map_err(map_eof)?;
+    let size = u32_as_usize(reader).map_err(map_eof)?;
```
VadimYarovoy, line 576: lets add after this lne next

```rs
// RFC 1813: the `count` field and the length of the opaque data describe
// the same payload, so a mismatch is a malformed request. Accepting one
// would let the backend read past the data it was handed.
if part_arg.size as usize != size {
    return Err(Error::IO(io::Error::new(
        ErrorKind::InvalidData,
        "WRITE count does not match opaque data length",
    )));
}
```

nfs-mamont/src/parser/read_buffer.rs
```diff
@@ -1,362 +1,190 @@
...
-                        )));
-                    }
-                    self.retry_mode = false;
-                    return Ok(val);
-                }
-                Err(err) => return Err(err),
-            }
+    /// `n` must not exceed the buffer capacity.
+    async fn ensure_buffered(&mut self, n: usize) -> io::Result<()> {
+        debug_assert!(n <= self.buf.len());
```
VadimYarovoy, line 91: why not error?

```rs
if n > self.buf.len() {
    return Err(io::Error::new(ErrorKind::InvalidData, "Read exceeds buffer capacity"));
}
```

nfs-mamont/src/parser/parser_struct.rs
```diff
@@ -687,123 +694,22 @@ where
...
     alloc: &Arc<A>,
-    buffer: &mut CountBuffer<S>,
+    reader: &mut FrameReader<S>,
 ) -> Result<(vfs::read::Args, A::Buffer)>
 where
     A: Allocator,
     S: AsyncRead + Unpin,
 {
-    let mut args = buffer.parse_with_retry(read::args).await?;
+    let args = read::args(reader).map_err(map_eof)?;
```
VadimYarovoy, line 653: and let's allow short read by

```rs
let mut args = read::args(reader).map_err(map_eof)?;

let count = min(args.count as usize, alloc.capacity().get());
args.count = count as u32;
```

VadimYarovoy (changes requested):
Some things to do and to think about

nfs-mamont/src/parser/parser_struct.rs
MikhailSvirukov: Probably we only need empty in case of 0-read request. Currently i suggest we hold to semantic when we allocate exact requested size or fail (without partial). If `allocate` fail there is probably some error, that affects whole server, so probably no point continue processing request

nfs-mamont/src/parser/parser_struct.rs
MikhailSvirukov: Let's do it in separate pr

nfs-mamont/src/parser/parser_struct.rs
VadimYarovoy, line 576: TODO for @VadimYarovoy - look rfc

nfs-mamont/src/parser/parser_struct.rs
MikhailSvirukov: Don't understand. Nothing bad would happen in case of discard rest of message, when there are 0 bytes in buffer. So, original error would be returned.

I don't see, where that `read_frame_header` would happen

VadimYarovoy (dismissed):
perf is good

VadimYarovoy:
@MikhailSvirukov update branch and it can be merged

nfs-mamont/src/parser/read_buffer.rs
MikhailSvirukov: I believe, that it is an invariant of buffer state. If it is broken, than there is a problem with parser's logic. I think, that we are ok, since we statically know sizes of all requests (including header of read requests) and guarantee, that buffer is big enought

nfs-mamont/src/parser/parser_struct.rs
MikhailSvirukov: #308

nfs-mamont/src/parser/parser_struct.rs
MikhailSvirukov: #308

nfs-mamont/src/parser/parser_struct.rs
MikhailSvirukov: fix, just in case - though probably that is not likely to happen, since they are `usize` (not `u32`)

VadimYarovoy (approved):

